PR #49, opened 2017-05-01
Introduce new rules from wikimedia/eslint-config-node-services

edg2s:
Includes many common sense rules we already follow.

Omits rules that are specific to ES6 or server-side environments.

The following stylistic rules is introduced, but is already
mostly followed in Wikimedia code & CSS:
* no-floating-decimal: .5 -> 0.5

The rule that triggers the most on Wikimedia code is probably:
* no-implicit-globals: var myGlobal -> window.myGlobal


jdforrester (approved):

edg2s:
We should remove these rules from node-services as downstream duplicates.

.eslintrc.json
```diff
@@ -26,23 +26,50 @@
     "no-array-constructor": "error",
     "no-bitwise": "error",
     "no-caller": "error",
+    "no-catch-shadow": "error",
```
Krinkle, line 29: +1.

This is IE<8 specific, so seems redundant. However, given that general block-level variables don't exist in ES5 (only ES6) this one occurrence of block-level variables is probably confusing if we allow it, so best to keep socially enforced that variable names are unique within the entire function.

.eslintrc.json
```diff
@@ -26,23 +26,50 @@
...
     "no-constant-condition": [ "error", { "checkLoops": false } ],
     "no-empty": [ "error", { "allowEmptyCatch": true } ],
     "no-eval": "error",
+    "no-extend-native": "error",
```
Krinkle, line 33: We used to enforce this with JSHint using `freeze`. Good to have this back!

.eslintrc.json
```diff
@@ -26,23 +26,50 @@
...
+    "no-extra-bind": "error",
+    "no-extra-label": "error",
+    "no-floating-decimal": "error",
     "no-implicit-coercion": [ "error", { "string": true, "boolean": false, "number": false } ],
+    "no-implicit-globals": "error",
     "no-implied-eval": "error",
+    "no-label-var": "error",
     "no-loop-func": "error",
     "no-multi-spaces": "error",
+    "no-multi-str": "error",
```
Krinkle, line 43: (Just for the record.) This too was previously enforced by JSHint already.

.eslintrc.json
```diff
@@ -26,23 +26,50 @@
...
+    "no-shadow-restricted-names": "error",
     "no-spaced-func": "error",
+    "no-throw-literal": "error",
     "no-trailing-spaces": "error",
+    "no-undef-init": "error",
     "no-underscore-dangle": "error",
+    "no-unmodified-loop-condition": "error",
     "no-unneeded-ternary": [ "error", { "defaultAssignment": false } ],
+    "no-unused-expressions": "error",
+    "no-unused-vars": [ "error", { "args": "none" } ],
```
Krinkle: `no-unused-vars` is part of eslint:recommended and already currently enforced without exception. I'm not sure we should loosen it to no longer check unused arguments.

<http://eslint.org/docs/rules/no-unused-vars>

Krinkle (changes requested):
All good, except the loosening of `no-unused-vars`.

.eslintrc.json
edg2s: Yes - won't cause an error, but could cause confusion.

.eslintrc.json
edg2s: Fixed one occurrence in MW core.

.eslintrc.json
edg2s: Oops.

edg2s:
Fixed.

Krinkle (approved):

PR #55, opened 2017-05-04
test: Add tests for negative rules

### LINKED ISSUE #5: Add tests for disallow rules
(From comments in #3)

While we have a test fixture that ensures our coding style is not rejected, disabling all rules would still make that pass. We need another test that asserts certain failures would happen when requirements are not met.

This will need to be more than a simple fixture since it's important to assert how many errors were triggered, which ones, and where. ESLint has a Node API that we can use for this.

Krinkle:
* Rename test/index and fixture to 'valid'.

* Update 'npm test' command to run eslint on all js files.
  This means it now applies to both test/valid.js and test/fixtures/valid.js.
  Previously it only tested the latter.

* Create new test called 'invalid':
  - Checks coverage (like valid.js)
  - Checks that warnings are as expected (using ESLint CLIEngine,
    and comparing results to a results fixture).
  - New 'npm run build-invalid' command to easily gegenerate
    this file after we changing rules or tests.

Fixes #5


package.json
```diff
@@ -24,6 +25,7 @@
   "author": "Gaidarenko Oleg <markelog@gmail.com>",
   "license": "MIT",
   "devDependencies": {
+    "assert-diff": "^1.2.0",
```
Krinkle, line 28: Using assert-diff as otherwise the results of `assert.deepEqual` failing is next to useless.

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "ESLint config following Wikimedia code conventions.",
   "main": ".eslintrc.json",
   "scripts": {
-    "test": "eslint test/fixtures/*.js && node test/index.js"
+    "test": "eslint . && node test/valid.js && node test/invalid.js",
```
jdforrester, line 7: You can't `eslint .` because invalid.js is caught by this (and, reasonably, doesn't pass).

Krinkle: No, it can. It's intentionally done that way. Any violations are marked with `disable-next-line`. I just forgot one of them because I was using an older version of ESLint locally that didn't have `no-unsafe-negation` enabled by default yet.

jdforrester (approved):
Neat.

PR #57, opened 2017-05-20
Remove explicit ecmaVersion

### LINKED ISSUE #56: Remove ecmaVersion: 5
This triggers a bug in eslint where inline /* eslint-env es6 */ stops working. This means it is impossible to have ES5 & ES6 files in the same folder, as the only way to override is with a new .eslintrc.json.

edg2s:
Fixes #56


jdforrester (approved):
Oy.

PR #63, opened 2018-01-23
New rule: `max-len` - Enforce 100 character line length

joakin:
Ignore:

* lines with eslint directives as they usually go over 80 because of the
  directive
* ignore long URLs
* ignore strings as it hurts readability on test files, assertions, and
  other strings that are hard to break

See https://phabricator.wikimedia.org/T185295 for more context and discussion.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,6 +22,16 @@
     "key-spacing": [ "error", { "beforeColon": false, "afterColon": true } ],
     "keyword-spacing": "error",
     "linebreak-style": [ "error", "unix" ],
+    "max-len": [ "warn", {
+      "code": 100,
+      "tabWidth": 4,
+      "ignorePattern": "^// eslint-.+",
+      "ignoreUrls": true,
+      "ignoreComments": false,
+      "ignoreRegExpLiterals": true,
+      "ignoreStrings": true,
+      "ignoreTemplateLiterals": true
+    } ],
     "new-cap": [ "error", { "newIsCap": true, "capIsNew": false, "properties": true } ],
     "new-parens": "error",
     "no-alert": "error",
--- a/test/fixtures/invalid-es5.js
+++ b/test/fixtures/invalid-es5.js
@@ -78,6 +78,9 @@ var APP;
 		// eslint-disable-next-line no-new-wrappers
 		bar = new String( 'events' );
 
+		// eslint-disable-next-line max-len
+		// This is a ludicrously long comment that is so long that one can't help but wonder as to whether the writer is a fool or merely testing other developers.
+
 		// eslint-disable-next-line no-new
 		new APP.Example();
 
--- a/test/fixtures/invalid-results.json
+++ b/test/fixtures/invalid-results.json
@@ -88,13 +88,13 @@
 			"filename": "fixtures/invalid-es5.js",
 			"line": 58,
 			"column": 4,
-			"ruleId": "no-inner-declarations"
+			"ruleId": "no-loop-func"
 		},
 		{
 			"filename": "fixtures/invalid-es5.js",
 			"line": 58,
 			"column": 4,
-			"ruleId": "no-loop-func"
+			"ruleId": "no-inner-declarations"
 		},
 		{
 			"filename": "fixtures/invalid-es5.js",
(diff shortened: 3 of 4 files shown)


joakin:
This needs discussion in the phab task before merging

jdforrester (dismissed):

.eslintrc.json
```diff
@@ -21,6 +21,16 @@
     "key-spacing": [ "error", { "beforeColon": false, "afterColon": true } ],
     "keyword-spacing": "error",
     "linebreak-style": [ "error", "unix" ],
+    "max-len": [ "warn", {
+      "code": 100,
+      "tabWidth": 4,
+      "ignorePattern": "^/// .+|// (?!eslint-.+max-len)eslint-.+|QUnit.",
```
Krinkle: We should ignore all lines starting with `// eslint-`, not just the ones disable something other than `max-len`. 

It's also not entirely clear to me what the `///` pattern is supposed to ignore or match. We should add some tests for all this.

jdforrester: This is intended for lines with an inline // eslint… (which we mostly discourage).

test/fixtures/valid.js
```diff
@@ -1,4 +1,5 @@
 // Rule: linebreak-style
+// Rule: max-len
```
Krinkle: This is not a real test :)

Let's add a regular line of code that (with indentation) just about exceeds 100 chars. And ensure it passes with eslint-disable-next-line over it, and also add the expected failure to the invalid fixture.

We should also have a second test that confirms really long `QUnit.test()` lines are allowed without explicit disable rule, and that long `// eslint-` lines are allowed as well.

.eslintrc.json
Krinkle: Yeah, but it's seems it excludes `// eslint-disable-next-line max-len`, which seems odd. The line disabling eslint for `max-line` (among other reasons) is just as likely to itself exceed the limit as any other disable line. I think this may be trying to fix a bug that might not exist.

jdforrester: Maybe drop the ignorePattern for now.

Krinkle (changes requested):
Using [this demo ...](https://eslint.org/demo/#eyJ0ZXh0IjoiLyplc2xpbnQgbWF4LWxlbjogW1wiZXJyb3JcIiwgeyBcImNvZGVcIjogNTAgfV0qL1xuXG5cbi8vIGVzbGludC1kaXNhYmxlLW5leHQtbGluZSBtYXgtbGVuXG52YXIgZm9vID0gMTAwMDAwMDAwMCArIDEwMDAwMDAwMDAgKyAxMDAwMDAwMDAwICsgMTAwMDAwMDAwMDtcblxuLy8gZXNsaW50LWRpc2FibGUtbmV4dC1saW5lIG1heC1sZW5cbi8vIGVzbGludC1kaXNhYmxlLW5leHQtbGluZSBuby11bnVzZWQtdmFycywgbWF4LWxlblxudmFyIGJhciA9IDEwMDAwMDAwMDAgKyAxMDAwMDAwMDAwICsgMTAwMDAwMDAwMCArIDEwMDAwMDAwMDA7XG5cblxuZnVuY3Rpb24gcXV1eChhKSB7XG5cdHJldHVybiBhO1xufVxuXG5xdXV4KGZvbyk7XG4iLCJvcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsiZWNtYVZlcnNpb24iOjUsInNvdXJjZVR5cGUiOiJzY3JpcHQiLCJlY21hRmVhdHVyZXMiOnt9fSwicnVsZXMiOnsiY29uc3RydWN0b3Itc3VwZXIiOjIsIm5vLWNhc2UtZGVjbGFyYXRpb25zIjoyLCJuby1jbGFzcy1hc3NpZ24iOjIsIm5vLWNvbXBhcmUtbmVnLXplcm8iOjIsIm5vLWNvbmQtYXNzaWduIjoyLCJuby1jb25zb2xlIjoyLCJuby1jb25zdC1hc3NpZ24iOjIsIm5vLWNvbnN0YW50LWNvbmRpdGlvbiI6Miwibm8tY29udHJvbC1yZWdleCI6Miwibm8tZGVidWdnZXIiOjIsIm5vLWRlbGV0ZS12YXIiOjIsIm5vLWR1cGUtYXJncyI6Miwibm8tZHVwZS1jbGFzcy1tZW1iZXJzIjoyLCJuby1kdXBlLWtleXMiOjIsIm5vLWR1cGxpY2F0ZS1jYXNlIjoyLCJuby1lbXB0eS1jaGFyYWN0ZXItY2xhc3MiOjIsIm5vLWVtcHR5LXBhdHRlcm4iOjIsIm5vLWVtcHR5IjoyLCJuby1leC1hc3NpZ24iOjIsIm5vLWV4dHJhLWJvb2xlYW4tY2FzdCI6Miwibm8tZXh0cmEtc2VtaSI6Miwibm8tZmFsbHRocm91Z2giOjIsIm5vLWZ1bmMtYXNzaWduIjoyLCJuby1nbG9iYWwtYXNzaWduIjoyLCJuby1pbm5lci1kZWNsYXJhdGlvbnMiOjIsIm5vLWludmFsaWQtcmVnZXhwIjoyLCJuby1pcnJlZ3VsYXItd2hpdGVzcGFjZSI6Miwibm8tbWl4ZWQtc3BhY2VzLWFuZC10YWJzIjoyLCJuby1uZXctc3ltYm9sIjoyLCJuby1vYmotY2FsbHMiOjIsIm5vLW9jdGFsIjoyLCJuby1yZWRlY2xhcmUiOjIsIm5vLXJlZ2V4LXNwYWNlcyI6Miwibm8tc2VsZi1hc3NpZ24iOjIsIm5vLXNwYXJzZS1hcnJheXMiOjIsIm5vLXRoaXMtYmVmb3JlLXN1cGVyIjoyLCJuby11bmRlZiI6Miwibm8tdW5leHBlY3RlZC1tdWx0aWxpbmUiOjIsIm5vLXVucmVhY2hhYmxlIjoyLCJuby11bnNhZmUtZmluYWxseSI6Miwibm8tdW5zYWZlLW5lZ2F0aW9uIjoyLCJuby11bnVzZWQtbGFiZWxzIjoyLCJuby11bnVzZWQtdmFycyI6Miwibm8tdXNlbGVzcy1lc2NhcGUiOjIsInJlcXVpcmUteWllbGQiOjIsInVzZS1pc25hbiI6MiwidmFsaWQtdHlwZW9mIjoyfSwiZW52Ijp7fX19), we see that eslint-disable generally works, but it does trigger on the `eslint-disable-*` lines if they exceed the configured limit. So we'll need to either disable it for inline comments or use an `ignorePattern`.

(Try by removing line 7 from the demo.)

Krinkle:
TODO:
* Add tests.
* Simplify rule config per <https://gerrit.wikimedia.org/r/#/c/420941/>. No need to backport these legacy exclusions.

test/fixtures/invalid.js
```diff
@@ -75,6 +75,12 @@ var APP;
 		// eslint-disable-next-line no-new-wrappers
 		bar = new String( 'events' );
 
+		// eslint-disable-next-line max-len
+		bar = 'This is a long string that is indeed so long that it breaches the line length rules and thus would trigger a warning were it not for the over-ride.';
+
+		// eslint-disable-next-line max-len
+		// This is a ludicrously long comment that is so long that one can't help but wonder as to whether the writer is a fool or merely testing other developers.
```
Krinkle: I suspect one or both of these cases aren't working properly. If they were working (as in, they fail without the inline comment), then the build should've been failing given this branch doesn't current add the expected warnings to `invalid-results.json`.

For the string, it not failing without the comment makes sense, given the `ignoreStrings: true` rule. However, I'm unsure as to why the long comment is also passing.

Krinkle (dismissed):

Krinkle (approved):

Issue #59, opened 2017-06-01
Review new rules in eslint 4.0.0


edg2s:
As of v4.0.0-beta.0 the following new rules look useful:

switch-colon-spacing: https://github.com/eslint/eslint/commit/482d5720c7faee174b192b2a131ec5ab93ceff3c

for-direction: https://github.com/eslint/eslint/commit/0ef09ea071f7df0cfb3f76ddea9ad746cb619912

Krinkle:
* <http://eslint.org/docs/rules/switch-colon-spacing>
* <http://eslint.org/docs/rules/for-direction>

Yep. LGTM, to both.

jdforrester:
We'll bump the (minor because we're < 0) version when we do this.

edg2s:
We should also add `"semi-style": [ "error", "last" ]`

jdforrester:
Can we declare this resolved?

PR #65, opened 2018-03-17
Update ESLint to version 4

Krinkle:
We already use ESLint 4.9.0 in mediawiki/core, but this repo
was still using eslint 3.x in its own tests.

Looks like a couple have things have changed in ESLint that we
should be aware of.

* ESLint 4.0.0 added 'no-useless-escape' to eslint:recommends,
  which we violated at test/invalid.js:45:11.
  Fixed.

* Indention rules have changed slightly. Only affected cases where
  the purpose of the line was *not* about indentation, so just
  fix the fixture to reflect current rule.

* Simplify test fixture for `valid-jsdoc/@returns`.
  At some point between eslint 3.1.1 and 4.19.0, the logic between
  "disable-next-line" and the "valid-jsdoc" broke.
  The valid-jsdoc rule was improved to have better line numbers,
  and now it can't be disabled before a multi-line comment.
  Filed upstream: https://github.com/eslint/eslint/issues/10095.
  Workaround: Use single-line doc comment for this case.

* Update invalid-results.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,6 @@
   "license": "MIT",
   "devDependencies": {
     "assert-diff": "^1.2.0",
-    "eslint": "^3.1.1"
+    "eslint": "^4.9.0"
   }
 }
--- a/test/fixtures/invalid-results.json
+++ b/test/fixtures/invalid-results.json
@@ -21,7 +21,7 @@
 	},
 	{
 		"line": 24,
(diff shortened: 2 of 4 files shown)


jdforrester (approved):

jdforrester:
We should consider #59 now.

PR #75, opened 2018-07-03
New rule: `arrow-steps`

jdforrester:
Also re-write the test suites to have ES5 and ES6 files each, and rename the
files-as-runners from the files-as-test-objects for clarity. As a bonus, we
now test the no-useless-computed-key rule, previously skipped due to needing
ES6 testing.


test/fixtures/invalid-es5.js
```diff
@@ -92,7 +92,7 @@ var APP;
 		}
 
 		// eslint-disable-next-line no-trailing-spaces
-		options.spaces = 'foo';
```
Krinkle: This is meant to have a trailing space, right?

jdforrester: Whoops, yes, fixed.

Krinkle (approved):

PR #80, opened 2018-07-13
Changed rule: valid-jsdoc — Set preferred tags where there are synonyms

niedzielski:
Set preferred tags for all synonyms listed[0]. The Closure Compiler tags
are also supported by default but no aliases were noted[1].

All changes were copied from the Services' Node.js config[2] except
yield, which was added.

http://usejsdoc.org/#block-tags
https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#jsdoc-tags
https://github.com/wikimedia/eslint-config-node-services/blob/master/.eslintrc.js#L219-L238


jdforrester:
Dupe of #81 – which do you want? [Edit: No it wasn't, my mistake.]

niedzielski:
@jdforrester, this is similar! #81 is for JavaScript type documentation like `boolean`, `string`, and so forth but this is for JSDoc tags like `@prop` and `@class`. So, for example, this patch would concern the `@param` tag of  `@param {boolean} foo` and #81 would concern the `{boolean}` type. Let me know if anything's unclear and I'll add comments where needed!

jdforrester:
Ah, OK, will re-write your commit titles to make it clear. :-)

.eslintrc.json
```diff
@@ -97,7 +97,34 @@
...
     "unicode-bom": [ "error" ],
-    "valid-jsdoc": [ "error", { "requireParamDescription": false, "requireReturnDescription": false, "requireReturn": false, "prefer": { "returns": "return" } } ],
+    "valid-jsdoc": [ "error", {
+      "requireParamDescription": false,
+      "requireReturnDescription": false,
+      "requireReturn": false,
+      "prefer": {
+        "abstract": "virtual",
+        "augments": "extends",
+        "constructor": "class",
```
Krinkle: I don't think these are the same?

It's been useful to differentiate in documentation between the class (as a whole) and the specific method signature of the constructor. We do this for PHP code with Doxygen, and currently for (the majority of) our code with JSDuck, in which `@class` and `@constructor` are different concepts. If this relates to JSDoc, then we should postpone this until later. I would also expect JSDoc to be able to make the same distinction, though perhaps through different means.

Example:

* <https://doc.wikimedia.org/mediawiki-core/master/js/#!/api/mw.Uri>
* <https://doc.wikimedia.org/mediawiki-core/master/js/#!/api/mw.Title>

Krinkle (dismissed):

.eslintrc.json
niedzielski: Maybe we should use `@constructs` instead?

http://usejsdoc.org/tags-class.html
http://usejsdoc.org/tags-constructs.html

Krinkle: In the present, that would produce an error with JSDuck.

Our current subset of tags are documented at <https://www.mediawiki.org/wiki/Manual:Coding_conventions/JavaScript#Tags>, with some projects defining additional custom tags and/or using JSDoc instead.

Once <https://phabricator.wikimedia.org/T187672> is further along,  we can revisit the defaults, but not yet.

We could, however, make a jsdoc-specific mixin, similar to the one we just landed for QUnit.

You'd use it like this:

```
{
  "extends:" [
    "wikimedia",
    "wikimedia/jsdoc"
  ]
}
```

Or something like that.

Meanwhile, we can also then add a stricter set of rules (rather than just a safe default subset) for JSDuck, under `wikimedia/jsduck`.

niedzielski: If this is the only issue, how about I just omit `"constructor": "class",` for the time being and it can be added later when JSDuck is replaced ([T138401](https://phabricator.wikimedia.org/T138401))? If you're ok with that, I'll update these tags according to the docs you've linked to.

Thanks for sharing your note on wikimedia/jsduck. I would like to see better JSDuck usage of course but I personally would prefer to invest my own energy converting repos to JSDoc instead.

Krinkle: Sounds good!

test/fixtures/invalid-es5.js
```diff
@@ -244,9 +244,33 @@ var APP;
 	 * @param {regexp} j
 	 * @return {error}
 	*/
-	APP.typing = function ( a, b, c, d, e, f, g, h, i, j ) {
+	APP.JSDocTypes = function ( a, b, c, d, e, f, g, h, i, j ) {
 		return a || b || c || d || e || f || g || h || i || j;
 	};
+
+	/**
+	 * {@linkcode APP}
```
niedzielski, line 252: These two inline tags don't seem to be linted properly at this time.

niedzielski:
Ready for review.

niedzielski:
Bump.

test/fixtures/valid-es5.js
```diff
@@ -234,10 +234,37 @@
...
+	 * @extends String
+	 * @class
+	 * @constructor
+	 * @constructs
+	 * @const
+	 * @default
+	 * @desc Description.
+	 * @external String
+	 * @file Description.
+	 * @emits Event
```
Krinkle: According to Codesearch, there are no uses of `@emits` in WMF-deployed repo. Instead, we use `@fires`.

niedzielski: :+1: done.

Krinkle:
Squashed.

Krinkle (approved):

PR #81, opened 2018-07-13
Changed rule: valid-jsdoc — Set preferred cases for types

niedzielski:
This patch specifies that lowercase types are preferred for boolean,
number, string, undefined, and null primitives, and uppercase for
Objects, Arrays, Functions, Dates, RegExps, and Errors which matches the
typing convention on wiki:

https://www.mediawiki.org/wiki/Manual:Coding_conventions/JavaScript#Types


niedzielski:
@jdlrobson, I can't seem to add you to this review but please check it out!

jdforrester:
Nice improvement!

Krinkle (changes requested):
@niedzielski Could you add positive and negative tests for these as well?

niedzielski:
@Krinkle, sure. Tests added :+1:

Krinkle (approved):

PR #86, opened 2018-08-10
New rule: `qunit/require-expect` (Fixes #85)

### LINKED ISSUE #85: Add "never-except-zero" setting to qunit/require-expect
Now that platinumazure/eslint-plugin-qunit/issues/69 has landed, let's use it.

jdforrester:
Requires upgrading to eslint-plugin-qunit 3.3.0+ to provide our
desired configuration option, 'never-except-zero'.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   "author": "Gaidarenko Oleg <markelog@gmail.com>",
   "license": "MIT",
   "peerDependencies": {
-    "eslint-plugin-qunit": "^3.2.1"
+    "eslint-plugin-qunit": "^3.3.0"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -30,6 +30,6 @@
   "devDependencies": {
```
Krinkle, line 30: To avoid an error, we need to bump the required version in `peerDependencies` as well.

Krinkle (changes requested):

package.json
jdforrester: Whoops.

Krinkle (approved):

PR #91, opened 2018-08-16
Changed rule: `quotes` – Add the 'avoidEscape' option

jdforrester:
Proposed during review of wikimedia/eslint-config-node-services/pull/29.


Krinkle:
Would be good to confirm it still rejects (needless) use of double quotes like it did before. Seems we were missing that from the invalid case. Probably because it doesn't have a `no-` prefix. Will need to be added to the list of rules to need an invalid case.

jdforrester:
Done.

Krinkle (approved):
Thanks!

PR #93, opened 2018-08-17
QUnit: Add the appropriate `env` setting

jdforrester:


Krinkle:
If this works the way I think it does, we can remove the (superfluous) setting from the test fixture. Can you check?

jdforrester:
Done. Works, at least locally. ;-)

edg2s (approved):

Krinkle (approved):

PR #95, opened 2018-08-17
Rename principal rules file to wikimedia.json, now we have several

jdforrester:
Replaces #94 with amendments.


package.json
```diff
@@ -20,7 +20,7 @@
     "wikimedia"
   ],
   "files": [
-    ".eslintrc.json",
+    "wikimedia.json",
```
edg2s, line 23: so we keep the .eslintrc.json file but don't ship it, is that right?

Krinkle (approved):

package.json
Krinkle: Yeah, I was surprised why James left it. But I assume it is for testing/editing files within this repository itself (project-local).

jdforrester: Yes. Otherwise I'd have had to do a messier change which points the default config for the repo's own files to something other than the default.

PR #97, opened 2018-08-23
Add services sub-config

jdforrester:
Needs a test suite to be written, but that could be added later?

Also @Pchelolo raised the reasonable concern on https://github.com/wikimedia/eslint-config-node-services/pull/30 that adding `peerDependencies` isn't ideal. Most downstream users won't use sub-config; now they will get warnings about not having them loaded… Thoughts?


package.json
```diff
@@ -21,16 +21,21 @@
...
+    "qunit.json",
+    "services.json"
   ],
   "author": "Gaidarenko Oleg <markelog@gmail.com>",
   "license": "MIT",
   "peerDependencies": {
-    "eslint-plugin-qunit": "^3.3.0"
+    "eslint-plugin-qunit": "^3.3.0",
+    "eslint-plugin-jsdoc": "^3.2.0",
+    "eslint-plugin-json": "^1.2.0"
```
Pchelolo: Technically `eslint` itself is a peerDependency, so if we're specifying them - we need to specify all of them. However, it's really inconvenient - all the peerDependencies must be listed in all the users `package.json` and the  versions must match.  I'm wondering if it would be better to make these regular dependencies, so that clients can simply depend on the config.

Alternatively, we could list the peerDependencies here. and make a proxy package, like `wikimedia-eslint` that would work as an aggregator for peer dependencies.

I li ke the former solution for simplicity even though it's not entirely correct from a philosophical perspective.

test/testInvalid.js
```diff
@@ -4,11 +4,17 @@ var assert = require( 'assert-diff' ),
 	eslint = require( 'eslint' ),
 	fs = require( 'fs' ),
 
-	fixture5 = fs.readFileSync( __dirname + '/fixtures/invalid-es5.js' ).toString(),
-	fixture6 = fs.readFileSync( __dirname + '/fixtures/invalid-es6.js' ).toString(),
-	fixtureQUnit = fs.readFileSync( __dirname + '/fixtures/qunit/invalid-qunit.js' ).toString(),
 	config = require( '../wikimedia.json' ),
 	configQUnit = require( '../qunit.json' ),
+	configServices = require( '../qunit.json' ),
```
Pchelolo: I believe that's a copy-paste error.

test/fixtures/services/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+	"extends": "../../../services.json"
+}
```
Pchelolo, line 3: Why do we need this? can't we just load it directly?

test/testValid.js
```diff
@@ -1,27 +1,33 @@
 /* eslint-env node */
 var assert = require( 'assert' ),
 	fs = require( 'fs' ),
+	config = require( '../wikimedia.json' ),
+	configQUnit = require( '../qunit.json' ),
+	configServices = require( '../qunit.json' ),
```
Pchelolo: Idem

Pchelolo (commented):
Left some comments

edg2s:
I have some concerns about introducing this set of rules a subset, and would like more feedback from the front-end standards group, as it is quite different from [our existing coding conventions](https://www.mediawiki.org/wiki/Manual:Coding_conventions) (e.g. uses spaces vs tabs). If individual teams are using different styles I don't think they should be forced to change, but also I don't think it should codified in our linting utilities, as that will discourage a move towards a more unified coding style.

Pchelolo:
The services config is used all over the backend JS stack and we're trying to gradually bring it as close as possible to the Wikimedia rules. Obviously, some differences will remain, like `env: nodejs` is only needed for backend, but the goal is to make it super close. 

Re space vs tabs, https://www.youtube.com/watch?v=SsoOG6ZeyUI On the serious note, I think this one is. impossible to change by now and it will remain a difference since changing would mean loosing all the git blame history throughout all the backend JS code.

edg2s:
Specifically, codifying these rules as part of the "wikimedia" ruleset. Right now I think they are too divergent to be merged, and having it as a separate ruleset makes that clearer.

I think a lot of the rules (e.g. 4 spaces) would need to be moved into project-specific overrides, so that at least new projects would benefit from the consistent style.

edg2s:
> Re space vs tabs, https://www.youtube.com/watch?v=SsoOG6ZeyUI 

You're dumped!

edg2s:
> On the serious note, I think this one is. impossible to change by now and it will remain a difference since changing would mean loosing all the git blame history throughout all the backend JS code

Actually `git blame -w` ignores whitespace and is the default in most UIs (e.g. SublimeText's git plugin).

jdforrester:
> I have some concerns about introducing this set of rules a subset, and would like more feedback from the front-end standards group, as it is quite different from our existing coding conventions (e.g. uses spaces vs tabs).

I agree that there's been a really unhelpful fork in coding standards that was pushed by a couple of people a few years ago, but this is about knitting them back together.

edg2s:
> I agree that there's been a really unhelpful fork in coding standards that was pushed by a couple of people a few years ago, but this is about knitting them back together.

We should merge in rules that make sense, but we shouldn't standardise things we don't want to be standards going forward (e.g. spaces). Those should be left as project-specific overrides,

Pchelolo:
I wanna propose a very drastic solution.

Using 
```
extends: 'eslint-config-wikimedia'
env:
  node: true
  es6: true
rules:
  indent: off
  space-in-parens: off
  computed-property-spacing: off
```
I've got RESTBase codebase to ~700 issues out of 30k LOC, out of which 400 are automatically fixable and are mostly related to whitespacing. The major concern with changing the whitespacing is git blame, but with `-w` it kinda goes away, so should we just have a broad discussion about merging the 2 completely and adjusting the services codebases into `-wikimedia` config with an exception on overriding the `env`?

jdforrester:
>   indent: off
  space-in-parens: off
  computed-property-spacing: off

Do you mean instead of 'off' the current over-ride values instead? You don't just want to not care, you want to retain the current style (at least for now)?

edg2s:
Definitely don't turn them off, that leaves no style enforced. The rule should always be set to tabs in this repo. Individual projects can have a local override.

Pchelolo:
> Do you mean instead of 'off' the current over-ride values instead? You don't just want to not care, you want to retain the current style (at least for now)?

Sorry I did not make myself clear. The purpose of my experiment was to demonstrate how close we are on the coding style, no to turn off the linting rules. And that services code can change only a little to accommodate the almost merged config. 

Since this created some cofusion, I will make a PR to demonstrate what I actually propose.

jdforrester:
Work moved to #109.

Issue #108, opened 2018-10-03
Create custom rule for jQuery selectors (or querySelector / getElementById)


edg2s:
I'd like to warn against using selectors:
 `$( '.dom-node-i-just-created-but-didnt-bother-to-store-in-memory' )`

This is equivalent to using a global variable (e.g. window.elementsByClass['dom-node...']) and should only be used to pick up nodes written to the page by other apps.

One would have to add exceptions to a few specific init files where your app is attached to the DOM, and maybe a few inline exceptions, but it would be quite similar to our stylelint rule that bans ID selectors.

Here's an example of a bug created by such code: https://phabricator.wikimedia.org/T204749

edit: Obviously one couldn't write a regex to distinguish `$( stringSelectorVariable )` from `$( domNodeVariable )`, so we would just limit the rule to `$( 'actual strings' )` but that will still catch the vast majority of issues.

Issue #107, opened 2018-10-03
Lint against string parsing in jQuery


edg2s:
Using string parsing to build DOM nodes is slow:

https://jsperf.com/jquery-buildfragment-vs-dom/1

Issue #106, opened 2018-09-18
Consider rules from eslint-plugin-jquery


edg2s:
https://github.com/dgraham/eslint-plugin-jquery allows disabling certain jQuery functions where faster native alternatives are available.

For example we could disable `$.inArray` as all our compatible browsers support `[].indexOf`.

edg2s:
I also would like to add performance lints, like don't use string parsing to build DOM nodes:

https://jsperf.com/jquery-buildfragment-vs-dom/1

This one is a bit more radical:

I'd like to warn against using any selectors:
 `$( '.dom-node-i-just-created-but-didnt-bother-to-store-in-memory' )`

This is equivalent to using a global variable (e.g. window.elementsByClass['dom-node...']) and should only be used to pick up nodes written to the page by other apps.

One would have to add exceptions to a few specific init files where your app is attached to the DOM, and maybe a few inline exceptions, but it would be quite similar to our stylelint rule that bans ID selectors.

Here's an example of a bug created by such code: https://phabricator.wikimedia.org/T204749

edit: Obviously one couldn't write a regex to distinguish `$( stringSelectorVariable )` from `$( domNodeVariable )`, so we would just limit the rule to `$( 'actual strings' )` but that will still catch the vast majority of issues.

edit2: Filed as #107 & #108

jdforrester:
I'm in favour of both of these; @jdlrobson and @Krinkle just added a series of bespoke rules to MediaWiki core's rules to effect the first: https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/core/+/refs/heads/master/.eslintrc.json#14

Krinkle:
Yeah, the `no-restricted-properties` rules were a good way to get started quickly, but I don't think we should use them in the long-term. At least not directly, because, while they support a custom message for display purposes, they don't have an identifier to use with `eslint-disable`. As such, the only way to turn them off is to turn off all rules within `no-restricted-properties`, which looks confusing in the code, but also means it disables more than intended.

Abstracting it within a custom rule will make the toggling code more understandable, and allows selective enabling/disabling as-needed.

The `eslint-plugin-jquery` is free of dependencies, well-written, well-tested, and is also being looked after well. **`+1`** for adopting some of its rules in our preset.

Shall we compile a list here for which rules we will initially use?

edg2s:
Well that patch seems a good place to start:
* each
* grep (not in eslint-plugin-jquery)
* inArray
[...]

Issue #96, opened 2018-08-20
Consider relaxing one-var for let and const


ricordisamoa:
The `one-var` rule is currently set to `[ "error", "always" ]` requiring variables to be declared in single statements at the top of their respective scope.

However, as the [ESLint documentation](https://eslint.org/docs/rules/one-var) puts it,
> The single-declaration school of thought is based in pre-ECMAScript 6 behaviors, where there was no such thing as block scope, only function scope. Since all `var` statements are hoisted to the top of the function anyway, some believe that declaring all variables in a single declaration at the top of the function removes confusion around scoping rules.

In the following example it would make sense to combine `var a, c` at the top of the function scope, but that is hardly the case with `let`.

```javascript
function () {
	// `a` and `c` are `undefined` here
	// accessing `b`, `d` or `e` here would throw a ReferenceError

	var a;
	let b;

	// `a`, `b` and `c` are `undefined` here
	// accessing `d` or `e` here would throw a ReferenceError

	if ( condition ) {
		var c;
		let d;
	}

	// `a`, `b` and `c` are `undefined` here
	// accessing `d` or `e` here would throw a ReferenceError

	let e;

	// `a`, `b`, `c` and `e` are `undefined` here
	// accessing `d` here would throw a ReferenceError
}
```

The same applies to `const` declarations as well.

Block scoping and [Temporal Dead Zone](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#Temporal_Dead_Zone) make it both sensible and convenient to declare variables just before use, mistakes being easily caught by ESLint's own [`no-undef`](https://eslint.org/docs/rules/no-undef) and [`no-use-before-define`](https://eslint.org/docs/rules/no-use-before-define). Why would ECMAScript 2015+ code have to trade Wikimedia style compliance for readability?

Something like `[ "error", { "var": "always", "let": "consecutive", "const": "consecutive" } ]` could be used to enforce consistency, e.g. `let a, b` in place of `let a; let b`. Alternatively, the latter form may be made admissible or even mandatory.

jdforrester:
With the services sub-config (#97) we'll be enforcing const/let over var and never-ing `one-var`; that would fix this for you, albeit in a preset designed for back-end services rather than front-end code.

ricordisamoa:
Thank you @jdforrester. However, it looks like that config will also have a lot of unrelated differences.

jdforrester:
[...]

PR #109, opened 2018-10-03
Refactor code to split into multiple profiles

jdforrester:
Replaces #97


edg2s:
Does this need to be changed to reflect the proposal in https://github.com/wikimedia/eslint-config-wikimedia/issues/106#issuecomment-429054694 ?

jdforrester:
> Does this need to be changed to reflect the proposal in [#106 (comment)](https://github.com/wikimedia/eslint-config-wikimedia/issues/106#issuecomment-429054694) ?

Yes, it needs updating a little (and rebasing). Patches welcome. ;-) Otherwise I'll get to it soon.

jdforrester:
OK, this is close to being good to merge.

The `client`, `qunit`, and `server` profiles need tests written for them.

The cross-tabulations for `es5`, `es6`, `es2015`, `es2016`, `es2017`, and `es2018` profiles will need writing (for #96, #113, _etc._). As a bonus, we should be able to sanity-test automatically against https://github.com/mdn/browser-compat-data/tree/master/javascript perhaps (which would help with avoiding this breaking in future).

The proposed new `jquery` profile (and especially all the non-performant function bans, for #106, #107, #108) needs writing and being added to `client` by default, I think, but I'd prefer to do that after this lands rather than drag this out forever.

Once all that's done, we can probably declare this profile to be 1.0.

Input welcome, especially if you disagree. :-)

client.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": [
+    "./common.json"
+  ],
+  "env": {
+    "es5": true,
+	"browser": true
```
Krinkle: Mixed spaces and tabs.

server.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": [
+    "./common.json"
+  ],
+  "env": {
+    "es6": true,
```
Krinkle: I think we should document what minimum Node version the preset is aimed at. The ES6 grammar was first introduced in Node 4, with most of the actual classes/methods arriving in Node 6. I'd say this currently resembles a Node 6+ environment.

Maybe also document that if a project's minimum support version is lower, to extend `common` directly instead (treat as custom config, starting with the basic preset). If we encourage projects to extend the new preset instead, they'd have to disable numerous seemingly unrelated rules that don't work well in older environments. In the past this was sometimes a source of frustration for us with upstream presets, and it'd be nice to avoid that from the start. We could also revisit that later, but if it's not controversial, might be better to start with. Given that it is easier to add support later (drop the social restriction), than to remove support later (add a new social restriction).

test/test.js
```diff
@@ -0,0 +1,92 @@
+/* eslint-env node */
+/* eslint-disable no-console */
+var count, config,
```
Krinkle: Hm.. I think most of these are meant to be local to the forEach callback, right? E.g. `config` isn't used in the initialisation scope.

test/test.js
```diff
@@ -0,0 +1,92 @@
...
+	fs = require( 'fs' ),
+
+	validFixturesFile, validFixtures,
+	assert = require( 'assert' ),
+
+	invalidFixturesFile, invalidFixtures, testPositivesFailures,
+	expectedFailuresFile, expectedFailures,
+	engine, report, results, prevFilename, prevLine,
+	assertDiff = require( 'assert-diff' ),
+	eslint = require( 'eslint' ),
```
Krinkle: I don't think we've written it down yet for Node code, but I believe we've been following the Python-esque convention of grouping the import/require statements together as the first group (in our grunt tasks, build scripts, as well as in services). That also matches what most other organisations and open source projects follow (even for otherwise different coding styles).

All other variables can still be grouped logically, but the imports are typically their own group.

Krinkle (commented):
LGTM, a few minor nits.

server.json
jdforrester: Yes, right now client is common+browser+es5 (current MediaWiki support) and server is common+node+es6 (current Wikimedia production node server version). When we break out the language support files that'll be more explicit, I think. Will document as such.

jdforrester:
I've tested this in a few repos with `"eslint-config-wikimedia": "git://github.com/wikimedia/eslint-config-wikimedia.git#e601586",` and it seems to work as expected.

server.json
```diff
@@ -0,0 +1,17 @@
...
+  ],
+  "env": {
+    "node": true
+  },
+  "rules": {
+    "arrow-parens": "error",
+    "no-misleading-character-class": "error",
+    "no-new-require": "error",
+    "no-useless-computed-key": "error",
+    "one-var": 0,
```
edg2s: we should use strings or numbers, but preferably not a mixture: https://eslint.org/docs/user-guide/configuring#configuring-rules

so 0 -> "off"

edg2s (changes requested):
Looks good, minor quibble, but lets just get this done.

edg2s (approved):

ricordisamoa:
With the [files](https://docs.npmjs.com/files/package.json#files) field set this way, is the 'language' directory included by NPM?

jdforrester:
> With the [files](https://docs.npmjs.com/files/package.json#files) field set this way, is the 'language' directory included by NPM?

I fiddled this when we released to make it work. I hope. :-)

PR #110, opened 2018-10-11
Introduce eslint-plugin-jquery and create wikimedia/jquery ruleset

### LINKED ISSUE #106: Consider rules from eslint-plugin-jquery

edg2s:
Fixes #106

### DIFF
--- a/jquery.json
+++ b/jquery.json
@@ -1,0 +1,15 @@
+{
+  "globals": {
+    "$": false
+  },
+  "plugins": [ "jquery" ],
+  "rules": {
+    "jquery/no-each": "error",
+    "jquery/no-grep": "error",
+    "jquery/no-in-array": "error",
+    "jquery/no-is-function": "error",
+    "jquery/no-map": "error",
+    "jquery/no-proxy": "error",
+    "jquery/no-trim": "error"
+  }
+}
--- a/jquery.json
+++ b/jquery.json
@@ -4,11 +4,11 @@
   },
   "plugins": [ "jquery" ],
   "rules": {
-    "jquery/no-each": "error",
+    "jquery/no-each-util": "error",
     "jquery/no-grep": "error",
     "jquery/no-in-array": "error",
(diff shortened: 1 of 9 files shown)


Krinkle:
LGTM. I think we can postpone the decision on browser/nodejs vs es5/es6 presets for a later time and get this out first.

Krinkle (approved):

jdforrester:
I disagree.

Krinkle:
Hm.. well, I don't mind the overhead of a few extra rules that would otherwise be ignored and harmless.

But, I am noticing something that might become more inconvenient, which is that due to being part of the default preset, it means projects will have to install the jquery plugin as peer dependency when upgrading, which isn't the case for qunit, so we may want to avoid that.

I've drafted a proposal at https://github.com/wikimedia/eslint-config-wikimedia/issues/106#issuecomment-429054694 that could resolve this.

edg2s:
Amended, this now creates an optional wikimedia/jquery ruleset.

jdforrester:
This is going to be a nightmare to rebase #109 onto, but is pretty easy to rebase this onto that…

edg2s:
Well if you're volunteering...

jdforrester:
> Well if you're volunteering...

I am, but I've had not one byte of feedback on 109 yet. ;-P

edg2s:
I'll rebase them the other way if you merge this one.

jdforrester (approved):
Weren't we going to use eslint-plugin-jquery from the Wikimedia repo?

jdforrester:
Blocked on https://github.com/wikimedia/eslint-plugin-jquery/issues/5 AIUI.

jdforrester (approved):

PR #112, opened 2018-11-01
New rule: `max-statements-per-line`

### LINKED ISSUE #111: Set max-statements-per-line to 1
We should add a rule like this:

    "max-statements-per-line": [ "error", { "max": 1 } ],

There are very few situations where more than one statement per line might be a good idea (mostly very short callback functions) and I think even such code is just as good when written out over multiple lines.

We adopted this in VisualEditor easily:
* https://gerrit.wikimedia.org/r/c/VisualEditor/VisualEditor/+/470518
* https://gerrit.wikimedia.org/r/c/mediawiki/extensions/VisualEditor/+/470519

edg2s:
Resolves #111

### DIFF
--- a/test/fixtures/invalid-es5.js
+++ b/test/fixtures/invalid-es5.js
@@ -248,6 +248,7 @@ var APP;
 		return a || b || c || d || e || f || g || h || i || j;
 	};
 
+	/* eslint-disable max-statements-per-line */
 	/**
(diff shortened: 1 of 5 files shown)


jdforrester:
Test failures are real.

jdforrester (approved):

PR #114, opened 2018-11-19
Rule split cleanup

edg2s:
* Extend correct rules in test fixtures
* Fix invalid-results for client.json
* Remove unused and empty valid.json


jdforrester:
Hmm. I'd hoped that we could find a rule that's specific to `client` rather than delete the empty file, but yeah…

jdforrester (approved):

edg2s:
valid.json, not valid.js - seems to have been left in error after a rename?

jdforrester:
Oh! 😊 Yes.

PR #119, opened 2018-11-20
Install jQuery plugin via GitHub to avoid rule prefix

edg2s:
Ideally we would install via the scoped NPM package,
but this results in giving all the rules an @wikimedia
prefix.


jdforrester:
I've re-done this in #118 instead.

PR #128, opened 2018-12-07
jquery: Add jquery as env

ghost:
I think it will be better to use
```json
env: {
    "jquery": true
}
```
Instead of 
```json
globals: {
    "$": false
}
```
Docs:
https://eslint.org/docs/user-guide/configuring#specifying-environments

### DIFF
--- a/jquery.json
+++ b/jquery.json
@@ -1,9 +1,7 @@
 {
-  "globals": {
-    "$": false
-  },
   "plugins": [ "jquery" ],
   "extends": "plugin:jquery/deprecated-3.3",
+  "env": { "jquery": true },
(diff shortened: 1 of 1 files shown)


jdforrester:
We did this intentionally as the `jQuery` global is being removed in MediaWiki code in lieu of `$`.

ghost:
@jdforrester sorry then :(

jdforrester:
Totally understandable, it's an oddity and it's undocumented. Thanks for the suggestion though. :-)

edg2s:
More details here: https://phabricator.wikimedia.org/T208951

PR #138, opened 2018-12-14
Fix invalid rules

edg2s:
* Move no-implied-eval from common to client as
  setTimeout( 'string' ) is only available in browsers.
* Fix trigger for no-empty
* Fix trigger for no-trailing-spaces

### DIFF
--- a/client.json
+++ b/client.json
@@ -8,6 +8,7 @@
   },
   "rules": {
     "no-alert": "error",
-    "no-console": "error"
+    "no-console": "error",
+    "no-implied-eval": "error"
   }
 }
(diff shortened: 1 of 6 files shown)


Krinkle:
Timers are available as global functions in Node.js as well, same as in browsers!

Krinkle (dismissed):
Timers are available as global functions in Node.js as well, same as in browsers!

Krinkle:
I assumed it would also include `new Function(string)` which is another indirect way of doing eval, and does exist in Node.js. But, that's covered by the `no-new-func` rule. Which we, together with `no-eval` still have in common.

jdforrester (approved):

PR #139, opened 2018-12-14
Replace invalid-results.json files with --report-unused-disable-directives

### LINKED ISSUE #135: Use reportUnusedDisableDirectives to get rid of invalid-results.json files
Our current approach for checking that a rule is being triggered is to first check that no errors are reported, then in order to check each inline comment correspond to an actual error suppression, we run eslint with `allowInlineConfig` set to false, and compare the results to the `invalid-results.json` file.

I think if we just run our checks with reportUnusedDisableDirectives, then that should effectively provide the same check: that each listed rule is actually suppressing an error.

edg2s:
Fixes #135


edg2s:
This is on top of #138

jdforrester:
Manual rebase needed, sorry.

jdforrester (approved):

PR #141, opened 2018-12-14
Remove negative rule from valid.js

edg2s:


jdforrester:
I thought we had this because this explicitly was allowing the "right" pattern and wasn't breaking things?

edg2s:
The no- rule is already tested in invalid.js, so while this line demonstrates that some version doesn't trigger an error and passes eslint, that isn't really asserted against a list of "there should be a positive check for this rule" like the other rules in this file. The fact it can be deleted without breaking the test runner show this.

jdforrester (approved):

PR #144, opened 2018-12-16
server: Move es6 rules to es6.json, node rules to node.json

### LINKED ISSUE #134: Allow users to use wikimedia/server with later versions of ES
Currently wikimedia/server uses ES6. Using `[ "wikimedia/server", "wikimedia/language/es2015" ]` won't work because the not-es2015 rules from es6 won't get removed.

edg2s:
Fixes #134


test/test.js
```diff
@@ -10,22 +10,36 @@ var fs = require( 'fs' ),
...
+	var count, config, rules;
 
 	console.log( 'Testing the "' + profile + '" profile suite.' );
 
 	config = require( '../' + profile + '.json' );
 	validFixturesFile = __dirname + '/fixtures/' + profile + '/valid.js';
 	invalidFixturesFile = __dirname + '/fixtures/' + profile + '/invalid.js';
+	rules = config.rules || {};
+
+	if ( profile === 'server' ) {
```
Krinkle, line 28: Hm.. I'm probably missing something, but why is this needed? I'd expect that ESLint can read the server.json file as-is from the above <profile>.json require. If not, and we need a hardcoded list of rules (which indeed line 42 indicates), then why is there not an exiting block for the client profile?

edg2s: Because server.json doesn't contain rules, it's just common+node+es6. We don't want to re-test common rules here. I considered moving the test fixtures to node/, but what we are actually testing is node+es6.

Arguably we should introduce the same abstraction for client, as one may want to use something other than ES5.

edg2s: ping @Krinkle

Krinkle (dismissed):

Krinkle (dismissed):
Resolved merge conflict.

Krinkle (approved):

PR #149, opened 2019-01-07
Minor formatting change to Changelog

Krinkle:
I copied this style from qunitjs originally, and thought it was smart to use `##` instead of `=======` because it saved a line and looked more "markdown"-ish (despite both being widely supported).

Now, having done QUnit releasing recently, I realise this was a mistake because it means the very useful tool "git-changelog", which QUnit use, doesn't work as expected.

See <https://github.com/tj/git-extras>, available on Homebrew.

Removing the first heading and using `=======` for the remaining makes it work, the rest was already conforming (naturally). It does support any file by the name of /(Change|History).*/i, so that's fine as well.


Krinkle:
In a nut shell, what the tool does is:
* find the latest semver tag,
* grab the log entries since then,
* open your configured `EDITOR` with an amended version of CHANGELOG.md with the log inserted as bullet point under a new heading with date etc.

Then, you make any adjustments, close it, and commit as usual. It's optional and easy to do manually as before for anyone that prefers that, but this would allow me to try to use that for the next release.

jdforrester (approved):
If we prefer this style now, we should go change all the other libraries that use this style (stylelint-config, OOjs, OOUI, …)

PR #154, opened 2019-01-15
Enable `template-curly-spacing`

shreyasminocha:
Since we're apparently fans of spacing our brackets out, here we go.

### DIFF
--- a/language/es6.json
+++ b/language/es6.json
@@ -12,6 +12,7 @@
     "no-new-require": "error",
     "no-useless-computed-key": "error",
     "one-var": "off",
+    "template-curly-spacing": ["error", "always"],
     "vars-on-top": "off"
   }
 }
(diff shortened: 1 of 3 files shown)


language/es6.json
```diff
@@ -12,6 +12,7 @@
     "no-new-require": "error",
     "no-useless-computed-key": "error",
     "one-var": "off",
-    "vars-on-top": "off"
+    "vars-on-top": "off",
+    "template-curly-spacing": "error"
```
Volker-E: Rules are ordered alphabetically here.

edg2s:
The default option is "never", I assume you meant to use "always"?

https://eslint.org/docs/rules/template-curly-spacing#always

edg2s:
Adding spaces would seem consistent with the rest of the ruleset, however I can see an argument for not adding non-meaningful whitespace to a string where whitespace is meaningful:

e.g. it may be easier to see the spacing is correct in the first example:
`The ${speed} brown fox`
`The ${ speed } brown fox`

Krinkle:
Superseded by #168, in which we went for not having spaces around variable interpolation.

PR #165, opened 2019-03-22
Upgrade js-yaml sub-dependency from 3.12.1 to 3.13.0 for DoS fix

jdforrester:


jdforrester:
I slipstreamed this into #167.

PR #168, opened 2019-04-25
[ES6] Breaking: avoid string concatenation

niedzielski:
Favor template literals and avoid unnecessary string concatenation.
Also update test/test.js with the new rules and no other changes.

### DIFF
--- a/language/es6.json
+++ b/language/es6.json
@@ -12,6 +12,7 @@
     "no-new-require": "error",
     "no-useless-computed-key": "error",
     "one-var": "off",
-    "vars-on-top": "off"
+    "vars-on-top": "off",
+    "prefer-const": "error"
   }
 }
--- a/language/es6.json
+++ b/language/es6.json
@@ -11,6 +11,8 @@
     "no-misleading-character-class": "error",
     "no-new-require": "error",
     "no-useless-computed-key": "error",
+    "no-useless-concat": "error",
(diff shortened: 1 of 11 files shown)


test/fixtures/server/valid.js
```diff
@@ -17,4 +17,11 @@
 		b--;
 	}
 
+	// Rule: no-useless-concat
+	// eslint-disable-next-line no-unused-expressions
+	'ab';
+
+	// Rule: prefer-template
+	// eslint-disable-next-line no-unused-expressions
+	`a${global}`;
```
Krinkle, line 27: `{var}` or `{ var }`? I've always seen and used `{var}`, but the next file in the this uses spaced `{ var }`.

niedzielski: Fixed. Added template-curly-spacing.

language/es6.json
```diff
@@ -13,6 +13,7 @@
     "no-useless-computed-key": "error",
     "no-useless-concat": "error",
     "prefer-template": "error",
+    "template-curly-spacing": "error",
```
jdforrester, line 16: This is a "positive" rule (not starting with `no-`) that triggers a test failure if used wrongly and should be covered in the positiveFailures files, except we don't test language rules yet. Meh.

jdforrester (approved):

PR #169, opened 2019-04-25
Hygiene: add test as a precommit hook

niedzielski:
Once installed via `npm i`, `npm t` is always executed on `git commit`
to avoid pushing bad commits. Given CI checks, it's unnecessary, but
some may find it useful.

https://www.npmjs.com/package/pre-commit

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -104,6 +104,12 @@
         "concat-map": "0.0.1"
       }
     },
+    "buffer-from": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/buffer-from/-/buffer-from-1.1.1.tgz",
+      "integrity": "sha512-MQcXEUbCKtEo7bhqEs6560Hyd4XaovZlO/k9V3hjVUF/zwW7KBVdSK4gIt/bzwS9MbR5qob+F5jusZsb0YQK2A==",
+      "dev": true
+    },
     "callsites": {
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/callsites/-/callsites-3.0.0.tgz",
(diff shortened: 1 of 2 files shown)


niedzielski:
Obviously, this change didn't take long to write. I submitted it because I thought it was useful but if the maintainers don't want it, just close the PR. It won't bother me a bit!

Krinkle:
For myself, I've usually find pre-commit hooks more stagnating than helpful during development. That's quite possibly due to my own workflow being different from yours. I don't make commits to this repo very often, so I'd like to hear from others what they think.

My own workflow on most projects is that during development I've got a tab where I iterate with `npm test` (or some other kind of command, e.g. partial karma/PHPUnit run which can be slow by default otherwise), in a second tab regularly evaluating unstaged changes and staging them by iterating with `git add -p` (`git p` for short). And lastly, a small always-open window where I draft the commit message (auto-saved as `~/.next_commitmsg.txt`).

Then happy with the result, the staging tab is where I'll run `dogitcommit` (tabbed from `dog`) which is a shortcut to `git commit -F ~/.nextcommitmsg`. I can then rebase, squash, or amend this commit as much as I want without being forced to re-run the test suite which I find really breaks up the mental flow of what I'm doing at any one moment.

For non-wikimedia projects that do enforce such a hook, I have a alias of a git commit command that bypasses it. Unfortunately, there are still many other ways that the commit hook still gets triggered (e.g. on amend, and on interactive rebase), where it's quite hard to bypass without setting up aliases for everything.

I'd generally recommend to go the other way - encourage developers that want to work this way, to set up an alias that runs the tests as part of their workflow. For example:

```
  p = add -p
  co = checkout
  tnc = ! npm test && git commit
  tnam = ! npm test && git commit --amend
```

Then, for when you want to test and commit (or test and amend, when iterating on a gerrit changeset) you run `git tnc` instead of `git commit`.

Anyway, as said, for this particular project, I'd like to hear from others as it won't affect me much either way.

niedzielski:
I normally run a subset of tests for projects I'm familiar with. For example, during development I don't care at all about style-only lints. I want to stay focused on the real problem, not whitespace, and it's only at commit time that I want to consider those concerns. The pre-commit hook is a useful reminder to cross the t's and dot the i's.

If I either feel confident that my changes won't fail the tests or don't wish to bother with them, I run `git commit -n` to skip the pre-commit hooks.

edg2s:
I actually find this feature quite annoying when working in skins/Minerva. If I want to make a quick amendment to a commit message, or I've just changed documentation, I have to wait 10-15 seconds to have the SVG compression checked. If I remember to add -n that helps, but I usually don't because most other projects don't require it. I can't always use -n because that also skips the git-review hooks.

Closing for lack of support here.

PR #178, opened 2019-07-05
Fix ES language rules tree

### LINKED ISSUE #117: Tag next major release
Hi,

it is great to see the new configuration for linting es5/es6 js has landed in master 🙌.
Do you have a timeline for when you plan to make a new official release to npm? We would love to use this updated configuration in the Wikibase extensions.

I assume this next release would be a major version bump due to the changed usage?

### LINKED ISSUE #177: es6.json should not extend not-es2015.json
https://github.com/wikimedia/eslint-config-wikimedia/blob/26eff2ee2bde3f510d5fdfa70f1849a15ae4172d/language/es6.json#L2

This is non-sensical. ES2015 is the 6th edition of ECMAScript. ES6 _**is**_ ES2015.

This causes issues when one wants to use the general "es6 rules" in `es6.json`, but, for example, in an es2017 environment that is then transformed by Babel to ES5. Because currently, one includes all the `no-restricted-syntax` rules from `not-es201*.json` when including `es6.json`. That shouldn't happen.

One way to model this properly could be to have a file `es6+.json` that includes the rules that should apply to all variations of ES6 javascript. This could then be extended by `es2015.json`, `es2016.json` and `es2017.json`.

edg2s:
* Merge ES6 rules into ES2015, and make ES6 an alias of ES2015
* Have language sets include the correct "not-" set, e.g. "not-2016" means syntax not allowed in ES2016.

Fixes #177


lucaswerkmeister:
Thanks already for the quick response!

> Fixes #117

#177 :)

~~I’m a bit confused about the not-* files now. If e. g. `es6.json` extends `not-es6.json`, then what’s the meaning of `not-es6.json`? I assumed it meant “features you can’t use because you’re **not** on **ES6** yet”, but now it sounds more like “the *not* part of `es6.json`”.~~ reading the PR description helps ^^

Also, in general, it would probably be useful if the `rules` part of `es6.json` was extracted into a separate file – then it could be included in e. g. `es2016.json` and `es2017.json` as well.

edg2s:
Thanks, fixed with new structure: https://github.com/wikimedia/eslint-config-wikimedia/blob/3f4c3c49c9e125e2858fd902787f2d436df76ec3/language/readme.md

lucaswerkmeister:
Thanks, that README helps a lot!

package.json
```diff
@@ -28,13 +28,17 @@
...
+    "language/rules-es6.json",
+    "language/rules-es2016.json",
+    "language/rules-es2017.json",
+    "language/not-es5.json",
     "language/not-es6.json",
-    "language/not-es2015.json",
     "language/not-es2016.json",
-    "language/not-es2017.json"
+    "language/not-es2017.json",
+    "language/es2015.json"
```
jdforrester, line 41: Maintain the sort?

jdforrester (approved):

package.json
edg2s: I put 2015 at the bottom because it's an alias

jdforrester: Fair.

PR #181, opened 2019-07-08
Fix language rule merge when key is unset in both rulesets

edg2s:
This fixes not-es2016.js where this occurs.

### DIFF
--- a/language/merge.js
+++ b/language/merge.js
@@ -1,10 +1,13 @@
 module.exports = function ( childRules, parentRules ) {
-	// Start off with a standard merge
-	const mergedRules = Object.assign( {}, parentRules, childRules );
+	// Only rules are preserved
+	const mergedRules = { rules: {} };
+
+	// Merge rules
+	Object.assign( mergedRules.rules, parentRules.rules, childRules.rules );
 
 	// For the specified keys, concatenate the lists
 	[ 'no-restricted-syntax', 'no-restricted-properties' ].forEach( function ( key ) {
-		if ( !mergedRules[ key ] ) {
+		if ( !mergedRules.rules[ key ] ) {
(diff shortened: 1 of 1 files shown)


edg2s:
We should release this as 13.1

jdforrester:
Test failures are sort-of real. Eurgh.

language/merge.js
```diff
@@ -1,18 +1,24 @@
...
+	Object.assign( mergedRules.rules, parentRules.rules, childRules.rules );
 
 	// For the specified keys, concatenate the lists
 	[ 'no-restricted-syntax', 'no-restricted-properties' ].forEach( function ( key ) {
-		// If either is unset, return the other
-		if ( !parentRules.rules[ key ] ) {
+		if ( !mergedRules.rules[ key ] ) {
+			// If both are unset, do nothing
+			return;
+		} else if ( !parentRules.rules[ key ] ) {
```
Krinkle: These to no-parent and no-child branches seem redundant, if I understand correctly, the below assignment results in the same outcome as the main merge above already did (a no-op). Is that right? Maybe I'm missing something...

If so, then these two branches could be removed, with the last `else` becoming an `else if` for `parent && child`. Alternatively, it looks like maybe that branch would work fine unconditionally as well, given it already has fallbacks for both of them being absent. Those fallbacks are currently inaccessible given these two earlier branches.

Krinkle (changes requested):
Either the inline fallbacks or the two extra branches seem redundant.

language/merge.js
edg2s: Yeah - it wasn't redundant when I didn't hardcode "error". Fixed.

Krinkle (approved):

PR #194, opened 2019-09-20
es6: Remove prefer-template rule

### LINKED ISSUE #193: Reconsider prefer-template rule
This was introduced in #168, but I'm not sure its a great idea. For long messages where there are lots of variables this template syntax may makes sense, e.g.

`'Hello ${user}, welcome to Wikipedia'`
or
`'<a href="${url}">${label}</a>'`

However the former is handled by our messages system, and the latter should be using DOM building.

More common cases on the server are path concatenations, in which case the extra syntax (`${}`) just makes the code harder to read if anything, compare:

```
__dirname + '/foo/bar'
```
with
```
`${__dirname}/foo/bar`
```

As a developer it is annoying that if I modify the above to remove the `__dirname` prefix, I have to change the string escape system. This also results in less clear diffs:
image
image

I don't think we should disallow templates, there will be cases with long strings and lots of variables where it makes sense, but I think enforcing this as a rule no matter how short the string is is not an improvement.

edg2s:
Fixes #193

### DIFF
--- a/language/rules-es6.json
+++ b/language/rules-es6.json
@@ -6,7 +6,6 @@
     "no-new-require": "error",
     "no-useless-computed-key": "error",
     "no-useless-concat": "error",
-    "prefer-template": "error",
     "template-curly-spacing": "error",
     "one-var": "off",
     "vars-on-top": "off",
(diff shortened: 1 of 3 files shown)


Krinkle (approved):

jdforrester (commented):
I'd like **some** input from the teams that said they wanted this before we just remove it. :-)

edg2s:
I think we've waited long enough? Let's merge this is and cut the next release (which would include this and the no-jquery updates).

PR #196, opened 2019-09-30
Update no-jquery rules

edg2s:
* Remove rules that in the deprecated set
* Add `no-error`


edg2s:
This depends on the next version of no-jquery

jdforrester (approved):

PR #209, opened 2019-10-25
mediawiki: New rule: `mediawiki/valid-package-file-require`

jdforrester:
Pull in eslint-plugin-mediawiki v0.2.1, up from v0.1.0.

### DIFF
--- a/mediawiki.json
+++ b/mediawiki.json
@@ -1,9 +1,11 @@
 {
   "globals": {
-    "mw": "readonly"
+    "mw": "readonly",
+	"require": "readonly"
   },
   "plugins": [ "mediawiki" ],
   "rules": {
-    "mediawiki/msg-doc": "error"
+    "mediawiki/msg-doc": "error",
+    "mediawiki/valid-package-file-require": "error"
   }
 }
--- a/mediawiki.json
+++ b/mediawiki.json
@@ -1,7 +1,7 @@
 {
   "globals": {
     "mw": "readonly",
-	"require": "readonly"
+    "require": "readonly"
   },
   "plugins": [ "mediawiki" ],
   "rules": {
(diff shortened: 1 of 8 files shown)


server.json
```diff
@@ -3,5 +3,9 @@
     "./common.json",
     "./node.json",
     "./language/es6.json"
-  ]
+  ],
+  "plugins": [ "mediawiki" ],
+  "rules": {
+    "mediawiki/valid-package-file-require": "error"
```
edg2s: This rule is for ResourceLoader requires, so should go in client, not server

edg2s (changes requested):

server.json
edg2s: Or rather, mediawiki.json

mediawiki.json
```diff
@@ -1,9 +1,11 @@
 {
   "globals": {
-    "mw": "readonly"
+    "mw": "readonly",
+	"require": "readonly"
```
edg2s: These files are still using space indentation :/

test/fixtures/mediawiki/invalid.js
```diff
@@ -3,4 +3,7 @@
 
 	// eslint-disable-next-line mediawiki/msg-doc
 	mw.msg( 'foo-' + bar );
+
+	// eslint-disable-next-line mediawiki/valid-package-file-require
+	require( 'foo' );
```
edg2s: this is valid - if you aren't getting an unused directive error that is a problem...

edg2s (changes requested):

test/fixtures/mediawiki/invalid.js
edg2s: Turns out the plugin includes a file-exists check, so we need to create "foo.js" then include it incorrectly to make it fail.

edg2s (approved):

PR #211, opened 2020-01-15
Replace deprecated valid-jsdoc with eslint-plugin-jsdoc

### LINKED ISSUE #210: valid-jsdoc rule is deprecated
ESLint team [stated](https://eslint.org/blog/2018/11/jsdoc-end-of-life) they will no longer maintain JSDoc support upstream, suggesting [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc) as a replacement.

jdforrester:
Fixes #210.


catrope (commented):
Dropping support for `@constructor` sucks, but if `valid-jsdoc` is going away I guess we'll have to?

edg2s:
We should check dropping `@constructor` doesn't change the docs output too much.

Krinkle:
`@constructor` is required when documenting a constructor. If the class in question has no constructor, or it has a constructor that needs no description and takes no parameters, then it can be ommited for near-identical output.

jdforrester:
I don't think it's sane to fork eslint-plugin-jsdoc to allow us to support both `@class` and `@constructor` tags, however.

Krinkle:
Looking upstream to see if there's a way we can "simply" configure this and it looks like this is indeed configurable already, via `tagNamePreference`.

However it's quite counter-intuitive (and undocumented) to re-allow something that has an alias/preference to something else.

See also <https://github.com/gajus/eslint-plugin-jsdoc/issues/367>

The following should work:

```yaml
 class: 'class'
```

This replaces the default alias of `class: 'constructor'`.

If that doesn't work, then we may need to do the inverse as well:

```yaml
class: 'class',
constructor: 'constructor'
```

jdforrester:
> The following should work:
> 
> ```yaml
>  class: 'class'
> ```
> 
> This replaces the default alias of `class: 'constructor'`.
> 
> If that doesn't work, then we may need to do the inverse as well:
> 
> ```yaml
> class: 'class',
> constructor: 'constructor'
> ```

I tried that on Tuesday and the over-ride didn't work, unless I've forgotten.

jdforrester:
> > The following should work:
> > ```yaml
> >  class: 'class'
> > ```
> > 
> > 
> > This replaces the default alias of `class: 'constructor'`.
> > If that doesn't work, then we may need to do the inverse as well:
> > ```yaml
> > class: 'class',
> > constructor: 'constructor'
> > ```
> 
> I tried that on Tuesday and the over-ride didn't work, unless I've forgotten.

Yeah, re-tried, not possible. I'd rather we didn't fork upsteam just for this, but if you insist…

edg2s:
This is broken due to an up-upstream issue in ESLint that doesn't let us use settings keys that conflict with Object.prototype.

The fix is pretty trivial, but it will probably take a while for it to propagate into a version we can use.

Upstream: https://github.com/gajus/eslint-plugin-jsdoc/issues/537
Up-upstream: https://github.com/eslint/eslint/issues/13289

common.json
```diff
@@ -5,6 +5,51 @@
 	],
 	"env": {},
 	"globals": {},
+	"plugins": [
+		"jsdoc",
+		"json"
```
edg2s: merge conflict: "json" was dropped here as it's inherited from json/recommended

common.json
```diff
@@ -100,45 +145,17 @@
...
-				"Null": "null",
-				"object": "Object",
-				"array": "Array",
-				"function": "Function",
-				"date": "Date",
-				"regexp": "RegExp",
-				"error": "Error"
-			}
-		} ],
+		"valid-jsdoc": [ "off" ],
```
edg2s: Not required, this is the upstream value

common.json
```diff
@@ -100,45 +145,17 @@
...
+		"jsdoc/check-alignment": "error",
```
edg2s: these should either go at the end, or alphabetically

edg2s: or better still we can add jsdoc.json

edg2s:
Test are now passing use a patched ESLint, but DNM until this patch is merged upstream.

edg2s:
My workaround patch to eslint-plugin-jsdoc had been merged and released with 25.4.1. This requires the object key to be 'tag constructor'.

jdforrester:
This seems fine, but I worry roll-out might be bumpy.

edg2s:
LGTM. Have you tested against a large repo (mw-core or VE)?

jdforrester:
> LGTM. Have you tested against a large repo (mw-core or VE)?

Some issues; might be blockers:
* A number of our tags aren't supported (because they're in jsduck but not jsdoc)
* The `Mixed` type isn't recognised (jsdoc issue?), and local "external" types have to be re-done in the .eslintrc.json file, but that's managable
* It doesn't like "indents" for continued text describing parameters/etc. which makes a serious mess of some of our VE stuff.
* Params of inherited things aren't picked up, which sucks greatly as we use that a lot in VE for inheritance of `config` and other params

Patch: https://gerrit.wikimedia.org/r/#/c/mediawiki/extensions/VisualEditor/+/596326

edg2s:
> * Params of inherited things aren't picked up, which sucks greatly as we use that a lot in VE for inheritance of `config` and other params

Use the `allowExtraTrailingParamDocs` option in `check-param-names` to fix this.

edg2s:
> * It doesn't like "indents" for continued text describing parameters/etc. which makes a serious mess of some of our VE stuff.

This can be fixed by adding `param`/`return` to the `excludeTags` option (in addition to `example`). This would still leave descriptions containing bullet lists (1 in VE, probably more in core), so we could just disable the rule, it doesn't catch a huge amount.

If we managed to fix the rule to allow indentation of all multi-line text blocks, the only errors we would catch would be `·*··@tag` -> `·*·@tag`.

Update: Should probably be disabled without https://github.com/gajus/eslint-plugin-jsdoc/issues/541. Lots of false positives in mw-core.

jsdoc.json
```diff
@@ -0,0 +1,80 @@
...
+	"settings": {
+		"jsdoc": {
+			"tagNamePreference": {
+				"alternateClassName": "alias",
+				"arg": "param",
+				"argument": "param",
+				"augments": "extends",
+				"cfg": "cfg",
+				"chainable": "chainable",
+				"context": "this",
```
edg2s, line 12: Do we need to choose one over the other for jsduck back compat?

Both are used across MW, maybe `@context` slightly more, but `@this` is the jsdoc standard so I went with that.

jsdoc.json
```diff
@@ -0,0 +1,79 @@
+{
+	"plugins": [ "jsdoc" ],
+	"settings": {
+		"jsdoc": {
+			"tagNamePreference": {
+				"alternateClassName": "alias",
```
edg2s, line 6: Is alternateClassName needed for jsduck compat? Otherwise we should use alias.

jsdoc.json
```diff
@@ -0,0 +1,56 @@
...
+				"tag constructor": "constructor",
+				"augments": "extends",
+				"constant": "const",
+				"defaultvalue": "default",
+				"description": "desc",
+				"emits": "fires",
+				"exception": "throws",
+				"fileoverview": "file",
+				"func": "function",
+				"function": "method",
```
edg2s, line 20: both func and function should map to method right?

edg2s:
> * local "external" types have to be re-done in the .eslintrc.json file, but that's managable

Only if they don't exist as globals in the code, which most of the time they do.

jsdoc.json
Krinkle: Sounds good to me. Note that neither `@context` nor `@this` is standard in JSDuck. It's part of our JSDuck config (the Ruby file) Updating to this version in repos will thus require a one-line change to that file, but that's no problem.

jdforrester:
OK, I think this is good to land:

https://gerrit.wikimedia.org/r/#/c/mediawiki/core/+/596509

jsdoc.json
edg2s: cc @Krinkle @jdforrester

jsdoc.json
edg2s: It used to be that "func" mapped to "function", which was very odd, because "function" mapped to "method", but the mapping was not transitive, so both method and function were allowed.

cc @Krinkle @jdforrester

edg2s:
I've dropped check-indentation as discussed. +1 from me if I can get a response on the outstanding inline questions above.

jsdoc.json
jdforrester: Yes.

jsdoc.json
jdforrester: Other than  jqXHR it seems rarely used; let's just ignore it: https://codesearch.wmflabs.org/search/?q=alternateClassName&i=nope&files=&repos=

jdforrester:
Good to land from my POV.

edg2s (approved):

PR #225, opened 2020-04-03
Add rules for .vue files

catrope:
Use the vue plugin for linting .vue files, and the es plugin to work
around the vue plugin's inability to support ES5-only environments.

Note that the Vue plugin does not report unused disable directives in
HTML templates, so the invalid.vue test files don't fully work. This
requires an upstream fix.

vue-common: Most Vue rules
vue-es6: ES6-specific Vue rules
vue-es5: ES5-specific Vue rules
vue-wrappers: Generate values for Vue rules that wrap core eslint
              rules (e.g. vue/eqeqeq)

### DIFF
--- a/mediawiki-vue.json
+++ b/mediawiki-vue.json
@@ -1,0 +1,11 @@
+{
+	"extends": [
+		"./vue-es5.js"
+	],
+	"rules": {
+		"no-implicit-globals": "off",
+		"vue/v-bind-style": [ "error", "longform" ],
+		"vue/v-slot-style": [ "error", "longform" ],
+		"vue/v-on-style": [ "error", "longform" ]
+	}
+}
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "ESLint config following Wikimedia code conventions.",
   "main": "common.json",
   "scripts": {
-    "test": "eslint . --ext .js,.json && node test/test.js",
+    "test": "eslint . --ext .js,.json,.vue && node test/test.js",
     "build-invalid": "node test/test.js dump"
   },
   "repository": {
@@ -25,8 +25,13 @@
     "node.json",
     "jquery.json",
     "mediawiki.json",
+    "mediawiki-vue.json",
     "qunit.json",
     "server.json",
+    "vue-common.json",
+    "vue-es5.js",
+    "vue-es6.js",
+    "vue-wrappers.js",
     "language/es5.json",
     "language/rules-es5.json",
     "language/not-es5.js",
@@ -57,10 +62,12 @@
   "license": "MIT",
   "dependencies": {
     "eslint": "^6.8.0",
+    "eslint-plugin-es": "^3.0.0",
     "eslint-plugin-json": "^2.1.1",
     "eslint-plugin-mediawiki": "^0.2.1",
     "eslint-plugin-no-jquery": "^2.3.2",
-    "eslint-plugin-qunit": "^4.0.0"
+    "eslint-plugin-qunit": "^4.0.0",
+    "eslint-plugin-vue": "^6.1.2"
   },
   "devDependencies": {
     "assert-diff": "^2.0.3"
--- a/readme.md
+++ b/readme.md
@@ -11,7 +11,7 @@ npm install --save-dev eslint-config-wikimedia
 Then, [configure ESLint](https://eslint.org/docs/user-guide/configuring) with one or more `.eslintrc.json` files as appropriate.
 
 ## Example configurations
-Bellow are some potential recommended uses:
+Below are some potential recommended uses:
 
 ### A typical front-end project
 This profile allows ES5 code and browser native functions. It will complain about ES6+ code and language features.
@@ -36,13 +36,30 @@ This profile adds the jQuery `$` global, and additional rules preventing the use
 ```
 
 #### MediaWiki
-Code that runs in MediaWiki can use this profile. It enforces rules that are specific to the MediaWiki codebase (core and extensions), such as correct documentation of `mw.message` usage.
+Code that runs in MediaWiki can use this profile. It enforces rules that are specific to the MediaWiki codebase (core and extensions), such as correct documentation of `mw.message` usage. It also automatically applies the Vue plugin and Vue-specific rules to `.vue` files, including MediaWiki-specific Vue rules such as prohibiting ES6 syntax and prohibiting shorthand syntax for `v-bind`, `v-on` and `v-slot`.
 `.eslintrc.json`:
 ```json
 {
 	"extends": [
 		"wikimedia/client",
 		"wikimedia/mediawiki"
+	]
+}
+```
+
+#### .vue files
+The MediaWiki profile automatically applies the Vue plugin to `.vue` files, and enforces MediaWiki-specific Vue rules. For code outside MediaWiki, or for `.vue` files that don't use ResourceLoader, you can use the `wikimedia/vue-es5` or `wikimedia/vue-es6` profile. These profiles only enforce non-MediaWiki-specific rules for Vue code.
+`.eslintrc.json`:
+```json
+{
+	"extends": [
+		"wikimedia/client"
+	],
+	"overrides": [
+		{
+			"files": "**/*.vue",
+			"extends": "wikimedia/vue-es5"
+		}
 	]
 }
 ```
--- a/readme.md
+++ b/readme.md
@@ -36,7 +36,7 @@ This profile adds the jQuery `$` global, and additional rules preventing the use
 ```
 
 #### MediaWiki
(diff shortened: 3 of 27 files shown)


vue-es5.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable quotes, quote-props  */
+module.exports = {
+	"extends": [
+		"./vue-es6.js",
+		"./language/not-es5.js",
```
edg2s, line 7: Why isn't this language/es5?

readme.md
```diff
@@ -47,6 +47,25 @@ Code that runs in MediaWiki can use this profile. It enforces rules that are spe
 }
 ```
 
+### .vue files in MediaWiki
```
edg2s: This should be a subsection of `#### MediaWiki`.

In general, eslint-config-wikimedia does not assume you are using MediaWiki.

Also if the rules are genuinely specific to using Vue in MW, then the rulesets should probably be named `mediawiki-vue*`.

package.json
```diff
@@ -60,9 +64,11 @@
...
     "eslint-plugin-mediawiki": "^0.2.1",
     "eslint-plugin-no-jquery": "^2.3.2",
-    "eslint-plugin-qunit": "^4.0.0"
+    "eslint-plugin-qunit": "^4.0.0",
+    "eslint-plugin-vue": "^6.1.2"
   },
   "devDependencies": {
-    "assert-diff": "^2.0.3"
+    "assert-diff": "^2.0.3",
+    "eslint-plugin-es": "^3.0.0"
```
edg2s: Isn't this a regular dependency?

readme.md
```diff
@@ -11,7 +11,7 @@ npm install --save-dev eslint-config-wikimedia
 Then, [configure ESLint](https://eslint.org/docs/user-guide/configuring) with one or more `.eslintrc.json` files as appropriate.
 
 ## Example configurations
-Bellow are some potential recommended uses:
+Below are some potential recommended uses:
```
edg2s, line 14: Spent too much time in Wales

vue-common.json
```diff
@@ -0,0 +1,60 @@
...
+				"LIFECYCLE_HOOKS",
+				"head",
+				[ "template", "render" ],
+				"renderError"
+			]
+		} ],
+		"vue/padding-line-between-blocks": [ "error", "always" ],
+		"vue/v-on-function-call": "error"
+	}
+}
```
edg2s: newline

test/fixtures/vue-wrappers/positiveFailures.json
```diff
@@ -0,0 +1,14 @@
...
+	"vue/camelcase",
+	"vue/comma-dangle",
+	"vue/dot-location",
+	"vue/eqeqeq",
+	"vue/key-spacing",
+	"vue/keyword/spacing",
+	"vue/object-curly-spacing",
+	"vue/space-infix-ops",
+	"vue/space-unary-ops"
+]
```
edg2s: newline

test/fixtures/vue-es5/positiveFailures.json
```diff
@@ -0,0 +1,5 @@
+[
+	"vue/v-bind-style",
+	"vue/v-on-style",
+	"vue/v-slot-style"
+]
```
edg2s: newline

test/fixtures/vue-common/positiveFailures.json
```diff
@@ -0,0 +1,4 @@
+[
+	"vue/v-on-function-call",
+	"vue/order-in-components"
+]
```
edg2s: newline

edg2s (changes requested):

vue-es5.js
jdforrester: Because Vue won't parse if ecmaScript is set to 5; it must be 6.

jdforrester:
In general, if this is something that is going to be default for all MediaWiki, it should be rolled into the `wikimedia/mediawiki` profile rather than require local configuration on hundreds of repos.

edg2s:
> In general, if this is something that is going to be default for all MediaWiki, it should be rolled into the `wikimedia/mediawiki` profile rather than require local configuration on hundreds of repos.

I'd rather no load a few hundred extra rules if you aren't even using vue.

jdforrester:
> > In general, if this is something that is going to be default for all MediaWiki, it should be rolled into the `wikimedia/mediawiki` profile rather than require local configuration on hundreds of repos.
> 
> I'd rather no load a few hundred extra rules if you aren't even using vue.

We're going to add it to almost every repo in the end. I'm not sure adding more friction is worth the few milliseconds of savings.

vue-es5.js
catrope: Now that this is a JS file (earlier on it was a JSON file) I can add a comment explaining this

package.json
catrope: Yes, oops, will fix

readme.md
catrope: Good point. Some of them are (like the no-v-bind rule, which is specific to ResourceLoader), some of them aren't. I'll try to disentangle this.

catrope:
> In general, if this is something that is going to be default for all MediaWiki, it should be rolled into the `wikimedia/mediawiki` profile rather than require local configuration on hundreds of repos.

Done. Code-wise, it's a very simple change (in `mediawiki.json`), so it would be easy to break out into a separate commit if we wanted. The only reason I didn't do that is because I would have had to write two almost completely different versions of the Vue section in `readme.md`.

edg2s:
> We're going to add it to almost every repo in the end. I'm not sure adding more friction is worth the few milliseconds of savings.

I'd rather wait and see if that is the case. We don't include jquery or client in mediawiki either.

jdforrester:
> > We're going to add it to almost every repo in the end. I'm not sure adding more friction is worth the few milliseconds of savings.
> 
> I'd rather wait and see if that is the case. We don't include jquery or client in mediawiki either.

For jquery, I thought that that was because we might want to reduce jquery usage, and that was a quick proxy for it? For client, yeah, we should have merged those in when we created the `mediawiki` profile, I think.

jdforrester (dismissed):
I'm in favour, but it sounds like Ed isn't, so let's not merge this until he is.

edg2s:
I'm not in favour, but I'm not as stubborn as you :)

vue-common.json
```diff
@@ -0,0 +1,60 @@
...
+		"vue/no-boolean-default": "error",
+		"vue/no-deprecated-scope-attribute": "error",
+		"vue/no-deprecated-slot-attribute": "error",
+		"vue/no-deprecated-slot-scope-attribute": "error",
+		"vue/no-reserved-component-names": "error",
+		"vue/no-static-inline-styles": "error",
+		"vue/no-unsupported-features": [ "error", {
+			"version": "2.6.11"
+		} ],
+		"vue/no-v-html": "off",
```
catrope: Reminder to self: remove this line

mediawiki-vue.json
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": [
+		"./vue-es5.js"
+	],
+	"rules": {
+		"no-implicit-globals": "off",
```
jdforrester, line 6: Is this really going to be needed for Vue? That's pretty crummy.

test/test.js
```diff
@@ -20,8 +21,12 @@ profiles.forEach( function ( profile ) {
 	console.log( `Testing the "${profileName}" profile suite.` );
 
 	config = require( `../${profile}` );
-	validFixturesFile = `${__dirname}/fixtures/${profileName}/valid.js`;
-	invalidFixturesFile = `${__dirname}/fixtures/${profileName}/invalid.js`;
+	validFixturesFile = fixtureExtensions
+		.map( ( ext ) => `${__dirname}/fixtures/${profileName}/valid.${ext}` )
+		.filter( fs.existsSync )[ 0 ];
```
jdforrester, line 26: This means you can't test both Vue and JS files in the same ruleset…

mediawiki-vue.json
catrope: It's not really needed, we could go without, but the no-implicit-globals rule is highly inconvenient for Vue files. As it is, it's already a bit silly especially in package modules, because RL wraps everything in closures already. In .vue files, the contents of the `<script>` tag get wrapped in a closure by RL (because `.vue` files can only be used in package modules), it doesn't make sense to force people to write IIFEs there.

test/test.js
jdforrester: Fixed in 0c33673c105532eac873f64e6a43653fae196c60.

mediawiki-vue.json
jdforrester: Fine.

edg2s: I think the rule also catches undeclared variables defined inside inner closures. The correct way to use this role would be to use sourceType:module, but that probably breaks other stuff.

catrope: eslint does not allow `sourceType: 'module'` to be combined with `ecmaVersion: 5`, because modules don't exist in ES5

catrope: However, if disabling `no-implicit-globals` also prevents us from catching other things, we might want to work around it in a different way, like avoiding top-level variables in .vue files wherever possible, and using `/* global foo */` where not.

jdforrester:
Let's follow-up in #228.

PR #232, opened 2020-04-16
mediawiki: Make vue rules only apply to vue files

jdforrester:


mediawiki.json
```diff
@@ -1,21 +1,45 @@
...
+  },
+  "overrides": [
+    {
+      "files": [ "**/*.vue" ],
+      "extends": [
+        "plugin:vue/recommended",
+        "./vue-es5.js"
+      ],
+      "plugins": [ "es" ],
+      "parserOptions": {
```
catrope: You shouldn't need to touch parserOptions here, vue-es5.js already does this

mediawiki.json
```diff
@@ -1,21 +1,45 @@
...
+        "rules": {
+          "vue/html-indent": [ "warn", "tab" ],
+          "vue/html-closing-bracket-newline": "off",
+          "vue/max-attributes-per-line": [ "warn", {
+            "singleline": 2,
+            "multiline": {
+              "max": 1,
+              "allowFirstLine": true
+            }
+          } ],
```
catrope: The rules above this point (html-indent, html-closing-bracket-newline, and max-attributes-per-line) should already be inherited from vue-common.json (which vue-es5.js indirectly inherits from)

catrope (changes requested):

mediawiki.json
jdforrester: Ah, yeah, was just hacking locally until it worked.

mediawiki.json
```diff
@@ -1,21 +1,30 @@
...
+    "mediawiki/valid-package-file-require": "error"
+  },
+  "overrides": [
+    {
+      "files": [ "**/*.vue" ],
+      "extends": [
+        "plugin:vue/recommended",
+        "./vue-es5.js"
+      ],
+      "plugins": [ "es" ],
```
catrope: This one is also duplicated from vue-es5.js, sorry for missing this the first time around

catrope (approved):

Issue #224, opened 2020-03-31
Use the es plugin rather than our own modelling of JS versions?


jdforrester:
Credit to @catrope for finding it. https://www.npmjs.com/package/eslint-plugin-es

edg2s:
Here's my findings so far:
* For all the static methods (`Object.values`, `Math.cosh` etc) it's an improvement. Each rule has it's own name so we don't have to worry about all the horrible hacking we are doing with `merge.js`
* It currently doesn't have any prototype methods, due to concerns over false positives. I've filed a task to at least include these as optional rules: https://github.com/mysticatea/eslint-plugin-es/issues/24
* All the other rules seem unnecessary if you have `ecmaVersion` set in ESLint as they only catch on things that already cause parser errors, e.g. `es/no-arrow-function` will never fire because you'll get an `unexpected token` error first. If we use this plugin we'll have to manually disable all these rules.

catrope:
> * All the other rules seem unnecessary if you have `ecmaVersion` set in ESLint as they only catch on things that already cause parser errors, e.g. `es/no-arrow-function` will never fire because you'll get an `unexpected token` error first. If we use this plugin we'll have to manually disable all these rules.

For my particular use case, this was a feature not a bug. `eslint-plugin-vue` crashes if `ecmaVersion` is set to 5, so I needed a way to prohibit ES6 features from being used while still setting `ecmaVersion` to 6.

I only did this for `.vue` files though, because I also found that it's much harder to disable these prohibitions again for code that is supposed to be ES6.

edg2s:
> this was a feature

Well it's a hack for a bug in another plugin. Also that sounds like a pretty bad bug, is it reported?

catrope:
Fair enough :) . I forgot to report the bug upstream yesterday, thanks for reminding me. https://github.com/vuejs/eslint-plugin-vue/issues/1089

PR #240, opened 2020-04-27
Use 'es' plugin in language/ where possible

edg2s:
Don't bother with syntax errors as these are caught
by the parser, but having individual rules for builtin
methods makes them easier to selectively disable.

Also ensure Array.prototype.entries rule doesn't conflict
with Object.entries.

Part of #224

### DIFF
--- a/language/not-es2016.js
+++ b/language/not-es2016.js
@@ -1,7 +1,11 @@
 /* eslint-disable quote-props, quotes */
 const merge = require( './merge.js' );
 const rules = {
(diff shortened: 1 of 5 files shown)


jdforrester (approved):
No upstream issue for padLeft/padRight etc.?

edg2s:
They're prototype methods

jdforrester:
> They're prototype methods

Ah, so https://github.com/mysticatea/eslint-plugin-es/issues/24 ? Sure.

PR #246, opened 2020-05-05
jquery: Update eslint-plugin-no-jquery to 2.4.1

edg2s:


edg2s:
This exposes issues that will be fixed in 2.4.1

jdforrester (approved):

PR #249, opened 2020-05-12
Update to ESLint 7.0.0

### LINKED ISSUE #248: Update to ESLint 7.0.0
This will be a breaking change obviously, but we already have several breaking changes (new rules) since 0.15.3.

jdforrester:
We can remove the --ext flag as these are evaluated
automatically now.

Fixes #248

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,9 +13,9 @@
       }
     },
     "@babel/helper-validator-identifier": {
-      "version": "7.9.0",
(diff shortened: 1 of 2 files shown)


edg2s:
I think you should review this as it's my commit.

jdforrester:
> I think you should review this as it's my commit.

Unfortunately I can't, because GitHub.

jdforrester:
(But consider this C+1'ed.)

edg2s (approved):

PR #253, opened 2020-05-15
Rule fix: Add balanced:true to spaced-comment

edg2s:
Enforces the whitespace at the end of a block comment.
Previously we only enforced at the start.


jdforrester:
How does this handle multi-line comments?

edg2s:
Example? We have /** doc comments and they appear unaffected

jdforrester:
> Example? We have /** doc comments and they appear unaffected

Yeah, `/**` won't be affected because of the `exceptions` block, but I thought that we had occasional multi-line ones which start `/*`; however, there's no matches for `\s\/\*[^!\*]+\n` in JS files in VE, MW or OOUI, at least, so never mind.

jdforrester (approved):

PR #260, opened 2020-05-25
Increase server.json ES verion to 2018

### LINKED ISSUE #259: Raise allowed ES version in server.json
Node 8 is EOL. As Node 10 has pretty good support for ES2018 (https://node.green/#ES2018) we should bump the version of ES that `server.json` uses.

edg2s:
Fixes #259

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -65,6 +65,12 @@
 				"uri-js": "^4.2.2"
 			}
 		},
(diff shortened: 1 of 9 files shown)


edg2s:
This is based on top of #256 so should be merged after that

jdforrester (approved):

PR #265, opened 2020-05-26
Require 'invalid' tests for every rule except ones set to 'off'

edg2s:


edg2s:
This is based on #260 so should be merged after that

jdforrester (approved):
This is great. Thank you.

PR #283, opened 2020-06-12
jsdoc: Extend from plugin:jsdoc/recommended

edg2s:
All errors are now warnings.
The following rules are now enabled:
* jsdoc/check-access
* jsdoc/require-param
* jsdoc/require-param-name
* jsdoc/require-param-type
* jsdoc/require-property-type
* jsdoc/require-returns
* jsdoc/require-returns-check
* jsdoc/require-returns-type

Assert all the rules of jsdoc/recommended in the test fixture.

### DIFF
--- a/jsdoc.json
+++ b/jsdoc.json
@@ -1,5 +1,5 @@
 {
-	"plugins": [ "jsdoc" ],
+	"extends": [ "plugin:jsdoc/recommended" ],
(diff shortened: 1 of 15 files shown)


edg2s:
Should be rebased after #282

jdforrester (approved):

PR #296, opened 2020-06-18
jsdoc: Drop @constant->@const, @description->@desc preferences

edg2s:
They are almost never used:
https://codesearch.wmflabs.org/search/?q=%5C%40const(%20%7C%24)&i=nope&files=.js&repos=
https://codesearch.wmflabs.org/search/?q=%5C%40desc(%20%7C%24)&i=nope&files=.js&repos=

and they are not jsduck builtins:
https://github.com/senchalabs/jsduck/wiki

### DIFF
--- a/jsdoc.json
+++ b/jsdoc.json
@@ -4,8 +4,6 @@
 		"jsdoc": {
 			"tagNamePreference": {
 				"augments": "extends",
-				"constant": "const",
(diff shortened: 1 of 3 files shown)


edg2s:
Merge after #295

jdforrester (approved):

PR #298, opened 2020-06-19
Vue: Prohibit self-closing tags, but allow shorthand attributes

catrope:
MediaWiki core has switched from libxml (which supports self-closing
tags, but breaks shorthand attributes) to RemexHtml (which does the
reverse).


catrope:
Should not be merged until https://gerrit.wikimedia.org/r/c/mediawiki/core/+/598133 is merged in MediaWiki.

jdforrester (approved):

PR #303, opened 2020-06-23
common: Enforce `no-unreachable-loop`

edg2s:
Fortunately we already some of these in the test fixtures.


edg2s:
Merge after #301

jdforrester (approved):

PR #305, opened 2020-06-24
common: Enforce `no-shadow`

### LINKED ISSUE #304: common: Consider no-shadow
https://eslint.org/docs/rules/no-shadow

edg2s:
Fixes #304

### DIFF
--- a/common.json
+++ b/common.json
@@ -69,6 +69,7 @@
 		"no-script-url": "error",
 		"no-self-compare": "error",
 		"no-sequences": "error",
+		"no-shadow": "error",
 		"no-shadow-restricted-names": "error",
 		"no-tabs": [ "error", { "allowIndentationTabs": true } ],
 		"no-throw-literal": "error",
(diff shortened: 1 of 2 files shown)


edg2s:
I would note that this rule can cause *lots* of warnings that are not always trivial to fix.

Krinkle (approved):

jdforrester (approved):

jdforrester:
We could start it as a warn if you're worried?

Krinkle:
Looks good to me.

I can see how it can cause warnings especially for increment variables and for lexical shadows of the same thing (such as `APP` in the test), where it's harmless.

I would recomend as such that we don't treat it as high priority to fix existing violations for projects where many exist today. eg. start by diabling it in eslintrc for those on most subdirectories and let it adopt slowly over time for older code, possibly as part of other refactors/cleanup, such as upcoming jsdoc changes.

PR #307, opened 2020-07-09
Update jsdoc plugin

lucaswerkmeister:
v29.1.3 of the jsdoc plugin fixes an issue when using the require-param rule (part of plugin:jsdoc/recommended) in --fix mode (gajus/eslint-plugin-jsdoc#607).

Note that this also pulls in the breaking release [v29.0.0](https://github.com/gajus/eslint-plugin-jsdoc/releases/tag/v29.0.0). Supposedly, that release would add spaces between union type items, i. e. turn `{string|number}` into `{string | number}`. However, I tried configuring this new version of the config in MediaWiki core, and it didn’t change any of the union types. I’m not sure if I’m doing something wrong or if there’s actually no change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -482,13 +482,13 @@
 			}
 		},
 		"eslint-plugin-jsdoc": {
-			"version": "28.5.1",
-			"resolved": "https://registry.npmjs.org/eslint-plugin-jsdoc/-/eslint-plugin-jsdoc-28.5.1.tgz",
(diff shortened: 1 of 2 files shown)


edg2s:
In the release notes it says "Removes JSDuck support (unions with "/" instead of "|")" - I assume that means that in JSDuck you could use "/" instead of "|", which we never did anyway.

edg2s (approved):

PR #318, opened 2020-09-07
common: Only run JSON rules on JSON files

edg2s:


edg2s:
This unbreaks our claimed support for dropping `--ext=js,json`, as non-standard extensions have to be included in `overrides` before they are searched for.

jdforrester (approved):

PR #319, opened 2020-09-07
Update ESLint from 7.6.0 to 7.8.1

edg2s:


package.json
```diff
@@ -49,7 +49,7 @@
 	],
 	"license": "MIT",
 	"dependencies": {
-		"eslint": "^7.6.0",
+		"eslint": "^7.8.0",
```
jdforrester: Do you want to actually pin to 7.8.1+? This is 7.8.0+.

edg2s: Yes, apparently I only pushed package-lock.json

jdforrester (approved):

PR #321, opened 2020-09-14
Update ESLint to 7.9.0

edg2s:

PR #322, opened 2020-09-14
qunit: Update eslint-plugin-qunit to 5.0.0

edg2s:
Added to /recommended:
* no-early-return (already enabled here)
* no-negated-ok (already enabled here)
* no-assert-logical-expresison (disabled)
* no-conditional-assertions (disabled)

We can consider enabling no-assert-logical-expresison &
no-conditional-assertions later, but the use cases in
MW don't seem obviously wrong.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -565,9 +565,9 @@
 			}
 		},
 		"eslint-plugin-qunit": {
-			"version": "4.3.0",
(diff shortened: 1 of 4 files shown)


edg2s:
Merge after #321

jdforrester (approved):

PR #336, opened 2020-12-07
Simplify language rules

edg2s:
Now that eslint-plugin-es-x contains rules for prototype properties,
we can do away with our not-es* rules and just rely on the plugin.


edg2s:
Depends on https://github.com/mysticatea/eslint-plugin-es/pull/54

jdforrester:
Exciting.

jdforrester:
The upstream was merged but still hasn't been released. Time to fork?

Krinkle:
> The upstream was merged but still hasn't been released. Time to fork?

Looks like someone already did. Worth taking a look perhaps. They've done quite a bit more than just releasing already-merged patches, however, so there's a certain risk of divergence here. Having said that, they seem to be keeping compatibility and mention explicitly an intent to discontinue the fork when maintenance resumes.

https://github.com/mysticatea/eslint-plugin-es/issues/72#issuecomment-1065820369

jdforrester (approved):

PR #337, opened 2020-12-09
json: Replace eslint-plugin-json with eslint-parser-json

edg2s:
This parses the JSON as JS, then applies ESLint style rules,
most importantly: indentation.

Unlike eslint-plugin-json this no longer checks the JSON
is completely valid, but in most cases an invalid JSON
file should cause your build to fail in another way.


edg2s:
Filed lack of JSON validation as https://github.com/zeitport/eslint-parser-json/issues/4

jsdoc.json
```diff
@@ -47,5 +47,5 @@
 		"jsdoc/require-property-description": "off",
 		"jsdoc/require-property-name": "off",
 		"jsdoc/require-returns-description": "off"
-    }
+	}
```
edg2s, line 50: yay

edg2s:
We can't just run eslint-plugin-json in addition to eslint-parser-json as the parser has been changed for all *.json files by the override.

jdforrester:
Interesting. Worth waiting on movement upstream?

edg2s:
Done

jdforrester (approved):

Issue #274, opened 2020-06-04
Consider supporting an ES6 client


edg2s:
This would requiring separating out the browser rules from `client`, perhaps to `browser`, then `client` would become `browser` + `language/es5`, and users could specify `browser` + `language/es6`.

Trying to do `client` + `language/es6` causes problems as language rules do not fully overwrite each other.

Note that this would support our current babel-transpiled projects as those are ES6 syntax but only support ES5 builtins (absent the use of plugins).

PR #358, opened 2021-03-10
Split client into client-es5 and client-es6

catrope:
Now that MediaWiki is about to support ES6 in client code, we need a
client ruleset for ES6.

- Move common client rules into client-common
- Remove vue-es5 extends from mediawiki, move it to client-es5
- Keep client as an alias for client-es5


edg2s:
This is issue #274

readme.md
```diff
@@ -15,55 +15,40 @@ Below are some potential recommended uses:
 
 ### A typical front-end project
 This config allows ES5 code and browser native functions. It will complain about ES6+ code and language features.
+It also automatically applies the Vue plugin and Vue-specific rules to `.vue` files.
```
edg2s, line 18: Are they Vue rules a generic Wikimedia style guide, or are they specific to how Vue is used in MediaWiki?

catrope: They are generic. The MW-specific rules were already in mediawiki.json, and I kept them there. A few things were specific to ES5 Vue code, and I kept those in vue-es5

catrope:
> This is issue #274

Thanks for pointing this out. I think the structure you lay out that issue is preferable (tell people to do `browser` + `language/es5` instead of `client-es5`, and `browser` + `language/es6` instead of `client-es6`), but I don't think that I can redo my PR with that structure, because there are rules that are specific to the combination of browser+ES5, and I don't think we could put those in either `browser` or `language/es5`.

edg2s:
Cool. I think `client-common` is an improvement on `browser` which is just a synonym of `client`.

readme.md
```diff
@@ -15,55 +15,40 @@ Below are some potential recommended uses:
...
 	"extends": [
-		"wikimedia/client",
+		"wikimedia/client-es5",
 		"wikimedia/jquery"
 	]
 }
 ```
 
 #### MediaWiki
-Code that runs in MediaWiki can use this config. It enforces rules that are specific to the MediaWiki codebase (core and extensions), such as correct documentation of `mw.message` usage. It also automatically applies the Vue plugin and Vue-specific rules to `.vue` files, including MediaWiki-specific Vue rules such as prohibiting ES6 syntax and prohibiting self-closing tags.
```
jdforrester, line 39: We should still mention the self-closing tag prohibition.

jdforrester (approved):

PR #361, opened 2021-03-18
Allow var blocks to be split and moved

### LINKED ISSUE #360: Change "one-var" requirement for es5 (and related rules)
In the ES6 preset, we turned off the `one-var` rule because the "one big comma-separated statement" style is uncommon in modern JavaScript and needlessly obfuscates code when writing or reviewing code. The reason it is uncommon is that `let` and `const` feature a "temporary dead-zone" in JavaScript, which means the caveat that motivated the "one-var" style, is only applicable to `var` and not other variable types like `let` and `const`.

I'm not sure why we kept the rule even for ES5 code, but I'm guessing it's because the underlying concern (variable hoisting) is of course still valid for actual `var` statements. However, I think this is really a very unlikely edge case to be worth keeping the rule for. The concern is basically to avoid code like this:

```js
function foo() {
 var x = 1;
 bar(y + 1);

 var y = 2;
}
```

If it weren't for style rules like this, the code would pass ESLint and the code would not fail with an (obvious) exception at run-time either, because the engine hoists var `y`, declared as `undefined`, to the start of the function. In ES6, when using `let` or `const` this is not a concern as those are not hoisted (leading to the "temporal dead zone", to ensure an error is thrown if you access them earlier).

The way we currently avoid this mistake is through the `one-var` rule. The intention being to bring the var statements together, so that you write code like this instead:

```js
function foo() {
 var x = 1,
     y = 2;
 bar(y + 1);
}
```

... or, if you are unable to compute `y` that early yet, then this will force you to realize there is a bug and deal with that, or to acknowledge that it is okay to use as `undefined` or some other default value, by declaring explicitly as such at the start:

```js
function foo() {
 var x = 1,
     y = 0;
 bar(y, 1);

 y = 2;
}
```

... Note that the `one-var` rule not only requires bringing the var statements together, it also currently enforces writing them as a single akwardly-indented comma-separated statement, and (to my surprise) I could not find anything at <https://eslint.org/docs/rules/one-var> that accmodates the style of "together, but as distinct var statements".

My guess is, the only reason it is like this historically is because it was easier to write such rule in JSHint, or that someone tried to optimise a few bytes in the output, or that simply nobody has enough recently about this rule to fix it.

### Proposal

I think the `one-var` rule is obsoleted by the [`no-use-before-define`](https://eslint.org/docs/rules/no-use-before-define) rule and so I'd like us to remove it, and allow developers to start writing code like such:

```js
function foo() {
 var x = 1;
 var y = 0;
 bar(y, 1);

 y = 2;
}
```

Existing code using the comma is totally fine, and may be preferred e.g. to group together variables that have no assignment:

```js
function foo() {
 var a, b, c;
 var x = 1;
 var y = 0;
 bar(y, 1);

 y = 2;
}
```

It does mean there's more than one way to write a statement, but I think this small amount of consistency will help authors to choose which style is most readable for the given code at hand, and also to slowly get used to this more modern coding style (if they want to).

My main motivation for this is that ESLint overrides and TypeScript declarations look less obtuse when they are not sandwhiched within a statement. As well as documentation comments more generally.

Compare:

```js
function foo() {
  var
    // Foo
    foo = 1,
    bar = 2;
}

function bar() {
  // Foo
  var foo = 1,
    // bar      <!-- indented, but not actually a logically nested block
    bar = 2;
}

function baz() {
  var /** @type {Object} */ foo = 1, // Help me
    // eslint-next-line-disable some-rule-here
    bar = 2;
}
```

To:

```js
function foo() {
  // Foo
  var foo = 1;
  var bar = 2;
}

function bar() {
  // Foo
  var foo = 1;
  // bar
  bar = 2;
}

function baz() {
  /** @type {Object} */
  var foo = 1;
  // eslint-next-line-disable some-rule-here
  var bar = 2;
}
```

edg2s:
* Disabled one-var and vars-on-top rules
* Enables block-scoped-var rule to prevent confusing usage.
  no-use-before-define should catch other problematic uses of var.

Fixes #360

### DIFF
--- a/common.json
+++ b/common.json
@@ -16,6 +16,7 @@
 	"rules": {
 		"array-bracket-spacing": [ "error", "always" ],
 		"array-callback-return": "error",
+		"block-scoped-var": "error",
 		"block-spacing": "error",
 		"brace-style": [ "error", "1tbs", { "allowSingleLine": true } ],
 		"camelcase": [ "error", { "properties": "always" } ],
@@ -94,7 +95,6 @@
 		"no-whitespace-before-property": "error",
 		"no-with": "error",
 		"object-curly-spacing": [ "error", "always" ],
-		"one-var": [ "error", "always" ],
 		"operator-linebreak": [ "error", "after" ],
 		"prefer-numeric-literals": "error",
 		"prefer-regex-literals": "error",
@@ -114,7 +114,6 @@
 		} ],
 		"switch-colon-spacing": [ "error", { "after": true, "before": false } ],
 		"unicode-bom": [ "error" ],
-		"vars-on-top": "error",
 		"wrap-iife": "error",
 		"yoda": [ "error", "never" ]
 	}
--- a/test/fixtures/common/invalid.js
+++ b/test/fixtures/common/invalid.js
@@ -4,7 +4,6 @@ var APP;
 ( function ( global ) {
 	// eslint-disable-next-line no-shadow
 	var APP;
-	// eslint-disable-next-line one-var
 	var upHere = function ( yArg ) {
 		var rArg = yArg.fooBar;
 		// eslint-disable-next-line semi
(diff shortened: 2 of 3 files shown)


jdforrester:
Seems good if we're doing this.

catrope:
Looks good to me. I verified that `for ( var key in obj ) { ... }` works too, as long as `key` is only used inside the loop.

Krinkle (approved):

edg2s:
Great, let's give the MW thread a week.

Issue #366, opened 2021-03-25
Consider qunit/no-assert-equal-boolean


edg2s:
This rules suggests `assert.true( foo )` over `assert.strictEqual( foo, true )`:
https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/no-assert-equal-boolean.md

Both check for a strict equality to `true`.

edg2s:
Personally I'm not sure `assert.true` adds much value. It is not immediately obvious that `assert.true` is strict.

jdforrester:
We did the equivalent of this for phpunit in mediawiki-codesniffer recently.

jdforrester:
(But that's with strict typing, as you point out.)

edg2s:
What happened in codesniffer?

edg2s:
The above patch will not enable this rule, but will disallow `assert.strictEqual( x > 5, true )` in favour of `assert.ok( x > 5 )`

jdforrester:
Shall we say that this is good enough and mark this as Resolved?

edg2s:
To argue with myself from 4 years ago:
* assert.true/false are [widely used anyway](https://codesearch.wmcloud.org/search/?q=assert.true&files=&excludeFiles=&repos=), as we don't have an inverse rule that lints against them
* The only we we can enforce consistency is with assert.true/false
* It doesn't really matter if people don't know that it is a strict comparison, it just matters that it is. Tests will fail if the value is not strictly true/false, and attempts to write the test another way will cause linting errors.

Personally I still think strictEqual( x, true ) is a bit clearer about what is going on behind the scenes, but given the above I don't object to us enforcing the other style.

PR #367, opened 2021-03-25
qunit: Update eslint-plugin-qunit to 6.0.0

edg2s:
Enables:
* no-async-module-callbacks
* no-compare-relation-boolean
* no-hooks-from-ancestor-modules
* no-nested-tests
* require-object-in-propequal

Disables:
* no-arrow-tests
* no-assert-equal-boolean (#366)


jdforrester (approved):
Will want to copy-paste the commit message for the change log.

PR #373, opened 2021-04-01
es5: Complain about use of Element.prototype.append()

jdforrester:


test/fixtures/client-es5/invalid.js
```diff
@@ -79,6 +79,8 @@
 	String.fromCodePoint();
 	// eslint-disable-next-line es/no-string-raw
 	String.raw();
+	// eslint-disable-next-line no-restricted-properties
+	document.append( node );
```
Volker-E: Why here and not underneath `document.appendChild()` on line 7?

Volker-E (commented):
See inline

language/not-es5.js
```diff
@@ -71,6 +71,10 @@ const rules = {
 			{
 				"property": "findIndex",
 				"message": "Unsupported method Array.prototype.findIndex requires ES6."
+			},
+			{
+				"property": "append",
+				"message": "Unsupported method Element.prototype.append requires ES6."
```
edg2s: This is DOM, so should go in client-es5.js

edg2s:
I'm not sure there's a neat way to stop this from triggering on jQuery.append :/ All the logic for detecting jQuery collections is tied up in the no-jquery plugin.

client-es5.js
```diff
@@ -14,6 +14,10 @@ const rules = {
 			{
 				"property": "parentElement",
 				"message": "Prefer parentNode to parentElement as Node.parentElement is not supported by IE11."
+			},
+			{
+				"property": "append",
```
Krinkle, line 19: Per Ed, this would conflict with non-DOM methods that are named "append". It's a pretty generic name. There's of course jQuery, but there's actually a fair number of other classes to consider as well: [Codesearch](<https://codesearch.wmcloud.org/deployed/?q=(%5E%7C%5C.%7C%5Cs)%5B%5E%24%5D%5Cw%2B%5C.append%5C(&i=nope&files=js%24&excludeFiles=jquery%7Cnode_modules&repos=>).

Results include uses in deployed code of Web APIs like URL SearchParams, BlobBuilder, TextRange, FormData, and Fetch Headers. And in our own source code (or vendored) through MobileFrontend View.js, `mw.translate`, D3, Ace InputStream, and more.

Krinkle (changes requested):

PR #376, opened 2021-04-07
Release v0.20.0

edg2s:


edg2s:
Merge after #375

jdforrester:
Do we want to pull in the new eslint-plugin-no-jquery first?

jdforrester:
Not before #377.

jdforrester (changes requested):
Need to mention the jquery change.

(Also, should we replicate the upstream things like "Rule fix: Add fixer for `no-on-ready`"?)

edg2s:
> (Also, should we replicate the upstream things like "Rule fix: Add fixer for `no-on-ready`"?)

I haven't done in the past. Unless it's significant, in which case it could be mentioned in the library update commit.

jdforrester (approved):
Ack.

PR #382, opened 2021-09-02
common: Remove no-multi-str rule

Krinkle:
The warning message reads:

> Multiline support is limited to browsers supporting ES5 only.

It existed to protect ES3 browsers per https://github.com/eslint/eslint/pull/133,
which we haven't needed to do for a while now.

Follows-up 53c67d32d2365, which copied this from node-services.

### DIFF
--- a/common.json
+++ b/common.json
@@ -63,7 +63,6 @@
 		"no-loop-func": "error",
 		"no-loss-of-precision": "error",
 		"no-multi-spaces": "error",
-		"no-multi-str": "error",
 		"no-multiple-empty-lines": [ "error", { "max": 1, "maxBOF": 0, "maxEOF": 0 } ],
 		"no-new": "error",
 		"no-new-func": "error",
--- a/test/fixtures/common/invalid.js
+++ b/test/fixtures/common/invalid.js
@@ -235,9 +235,6 @@ var APP;
 	APP.defaults =  {
 		// eslint-disable-next-line no-floating-decimal
 		decimal: .4,
-		// eslint-disable-next-line no-multi-str
-		multiStr: ' \
-',
 		/* eslint-disable no-multiple-empty-lines */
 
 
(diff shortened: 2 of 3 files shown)


edg2s:
1. Should we be encouraging template literals in ES6 instead?

2. Given that this has been part of ES5 styleguide for 4 years most of our codebase is consistently using:
```js
'multi' +
'line'
```
Is there much value in allowing another style?

Krinkle:
The rule was added here in 2016, which was quite late. Use of multi line strings was more common in the years prior, for HTML templates and `jquery.localize` etc. Most of that had been converted to OOUI widgets or Mustache templates by then.

There are still a few uses of it ([Codesearch](https://codesearch.wmcloud.org/deployed/?q=no-multi-str&i=nope&files=js%24&excludeFiles=node_modules&repos=)), but it's quite rare. And while each use has an eslint-disable rule above it, probably no-one has bothered to remove it from the preset as we rarely need them.

My thinking is that, the idea of multi-line strings themselves pose no problem for quality or maintainability. So on the rare occasion someone does find a use for it, should they be discouraged and/or be faced with a dated warning about ES3 support?

I'm writing a patch in mediawiki/core involving the following snippet:

```js
// eslint-disable-next-line no-multi-str
var code = "{\
	main: 'init.js',\
	files: {\
		'data.json': { hello: 'world' },\
		'alice.js': function ( require, module ) {\
			var core = require( './core.js' );\
			module.exports = core.sayHello( 'Alice' );\
		},\
		'core.js': function ( require, module ) {\
			module.exports = {\
				sayHello: function ( name ) {\
					return 'Hello ' + name;\
				}\
			};\
		},\
		'init.js': function ( require ) {\
			mw.alice = require( './alice.js' );\
		}\
	}\
}";
```

> Should we be encouraging template literals in ES6 instead?

I believe we should, yes. In purpose, that feels like part of `prefer-template`, although it doesn't currently warn on multi-line strings alone. It only warns if there are variables or other concatenations involved. We could augment it with `no-multi-str` in the es6-specific preset, but, it do note that its purpose and warning would still be confusing. We may want to instead improve `prefer-template` to detect multi-line strings and warn against them with a more actionable warning in ES6 code.

edg2s (approved):

PR #397, opened 2021-10-06
Add Vue 3 support

### LINKED ISSUE #393: Prepare for Vue 3
eslint-plugin-vue has rules specific to Vue 3. We should provide a Vue 2 config and a Vue 3 config, with the latter including the Vue 3 rules and setting `no-unsupported-features` to a different value.

catrope:
- Move version-specific things from vue-common to vue2-common and
  vue3-common
- Rename vue-es5 and vue-es6 to vue2-es5 and vue2-es6 (but keep old
  names as aliases)
- Add vue3-es6
- Move tests from vue-common to vue2-common, since vue-common is not a
  standalone ruleset anymore
- Point client-es5 and client-es6 to vue2-es5 and vue2-es6 respectively.
  We'll probably point client-es6 to vue3-es6 in the future.

Fixes #393


edg2s (changes requested):
I think README needs updating

edg2s (approved):
Actually, there's no references to the vue presets directly. We probably should document the vue2/vue3 stuff at some point though.

PR #411, opened 2021-10-28
mediawiki: Allow long message keys in comments

### LINKED ISSUE #399: mediawiki/msg-doc should suppress length warnings
`mediawiki/msg-doc` requires message keys to be enumerated in comments when a non-string argument is passed to `mw.msg()` or similar methods. Message keys tend to be long, and there is no possibility for line breaks in the comments. `max-len` warnings on the same line should be suppressed, if possible.

edg2s:
Fixes #399

### DIFF
--- a/mediawiki.js
+++ b/mediawiki.js
@@ -1,4 +1,9 @@
-{
+'use strict';
+
+const commonRules = require( './common' );
+
+/* eslint-disable quote-props, quotes */
+const config = {
 	"globals": {
 		"mw": "readonly",
 		"OO": "readonly",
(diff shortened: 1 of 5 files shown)


mediawiki.js
```diff
@@ -35,4 +40,15 @@
...
-}
+};
+
+// Extend max-len.ignorePattern
+config.rules[ 'max-len' ] = commonRules.rules[ 'max-len' ];
+config.rules[ 'max-len' ][ 1 ].ignorePattern =
+	'(' +
+		config.rules[ 'max-len' ][ 1 ].ignorePattern +
+		'|' +
+		'[\\s]*(//|<!--) \\* [a-z-]+$' +
```
catrope: I think we should also allow numbers and uppercase letters in message names. Why not use the same regex that msg-doc [uses internally](https://github.com/wikimedia/eslint-plugin-mediawiki/blob/master/src/utils.js#L13)? That one is much more permissive.

catrope (changes requested):

mediawiki.js
edg2s: Yes, we should allow numbers. IIRC upper case letters are not supposed to be used in message keys and can cause problems, although they have been used in some extensions erroneously so maybe we should allow them.

That pattern you link too is very loose because it allows things with spaces like "* label-type-0, label-type-1, etc.". For a line length rule we shouldn't be allowing spaces.

catrope (approved):

PR #415, opened 2021-10-29
Update eslint-plugin-vue to 8.0.2

### LINKED ISSUE #408: Consider new rules from eslint-plugin-vue 7.20.0
https://github.com/vuejs/eslint-plugin-vue/releases/tag/v7.20.0

catrope:
- Remove rules that we had manually enabled but are now part of the
  preset:
    - vue/no-useless-template-attributes
    - vue/valid-next-tick
    - vue/no-export-in-script-setup
    - vue/valid-define-emits
    - vue/valid-define-props
    - vue/valid-v-memo
    - vue/no-deprecated-v-is
- Add vue/no-undef-properties (new upstream in 7.20.0, not in preset)
- Update vue/max-attributes-per-line for breaking change
    - Adopt upstream standard for vue/first-attribute-linebreak, which differs
      from what we had previously allowed

Closes #408.


edg2s (approved):

edg2s:
By default, the test suite will only make assertions for rules which are set locally, however there is an option to assert for inherited rules too. We should probably enabled this for the vue config at some point: https://github.com/wikimedia/eslint-config-wikimedia/blob/master/test/test.js#L54

PR #420, opened 2021-11-02
vue: Assert upstream config settings

edg2s:
This is useful when upstream changes.

### DIFF
--- a/test/fixtures/vue2-common/invalid.vue
+++ b/test/fixtures/vue2-common/invalid.vue
@@ -17,6 +17,20 @@
 				{{ baz }}
 			</template>
 		</blah-component>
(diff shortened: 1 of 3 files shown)


edg2s:
TODO: There are >60 rules to add to `invalid.vue`

jdforrester:
Oh right, your TODO is a TODO-before-merge. :-)

edg2s:
> Oh right, your TODO is a TODO-before-merge. :-)

Yeah - maybe @catrope will have a go when he's bored, because he understands these rules better than me so could write better tests.

edg2s:
As this only for tests, it needn't block the release.

PR #423, opened 2021-11-02
qunit: Replace various rules with `no-loose-assertions`

edg2s:
assert.equal/ok/notOk and various special cases of them were
covered by assorted rules.

Replace all these rules with `no-loose-assertions` which
also disallows assert.notEqual.

### DIFF
--- a/test/fixtures/qunit/invalid.js
+++ b/test/fixtures/qunit/invalid.js
@@ -5,8 +5,11 @@ QUnit.module( 'Example' );
 QUnit.test( '.foo()', function ( assert ) {
 	const x = 'bar';
 
-	// eslint-disable-next-line qunit/no-assert-equal
+	// eslint-disable-next-line qunit/no-loose-assertions
 	assert.equal( x, 'bar' );
+
+	// eslint-disable-next-line qunit/no-loose-assertions
+	assert.notEqual( x, 'bar' );
 
 	assert.expect( 3 );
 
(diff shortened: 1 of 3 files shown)


test/fixtures/qunit/invalid.js
```diff
@@ -5,21 +5,24 @@ QUnit.module( 'Example' );
...
+
 	assert.expect( 3 );
 
 	if ( !x ) {
 		// eslint-disable-next-line qunit/no-early-return
 		return;
 	}
 
-	// eslint-disable-next-line qunit/no-assert-ok
-	assert.ok( x );
```
Krinkle, line 19: Did you mean to remove this one? Plain use of `ok()` seems like it'd still be useful to warn against (more common than the negated one we still test for). I think `no-loose-assertions` warns for both.

edg2s: Yes - I figured we only needed one, and leaving the negated one sort of asserts that we don't get the no-negated-ok error as well.

test/fixtures/qunit/valid.js
```diff
@@ -2,25 +2,31 @@ QUnit.module( 'Example' );
 
 // Off: qunit/no-arrow-tests
 // Valid: qunit/require-expect
-QUnit.test( '.foo()', ( assert ) => {
+QUnit.test( '.foo()', () => {
+} );
+
+QUnit.test( '.bar()', function ( assert ) {
```
edg2s, line 8: I moved most of our code out the arrow function, because most eslint-plugin-qunit rules are broken in arrow functions (and always have been), see https://github.com/platinumazure/eslint-plugin-qunit/issues/213

Krinkle (approved):

PR #426, opened 2021-11-24
common: Add eslint-plugin-yml

### LINKED ISSUE #425: Add linting for YAML files
Originally filed as https://phabricator.wikimedia.org/T296367 by @Pchelolo. https://www.npmjs.com/package/eslint-plugin-yml indeed looks plausible.

Pchelolo:
The rules available to us are fairly simple. I've modified
a few rules to respect wikimedia more-spaces-are-better
policy. You can see the examples of some of the invalid
YAMLs in the tests.

One large difference with other file types is that YAML
explicitly prohibits the use of tabs as indentation, so
we standadise on 2 spaces.

Fixes #425


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "eslint-config-wikimedia",
-	"version": "0.21.0",
+	"version": "0.22.0",
```
jdforrester: This isn't a release commit; these are bumped in release commits only.

jdforrester:
Eurgh, GH sucks so very badly.

See my trivial fix for your PR which I can't apply to this one without creating a broken (non-squashed) git entry by merging yours first, as you did the "right" thing and made this a PR from your fork rather than a local branch. Meh.

package.json
Pchelolo: Fixed

test/fixtures/yaml/invalid.yaml
```diff
@@ -0,0 +1,17 @@
+object:
```
edg2s, line 1: We mostly use eslint-disable-next-line when possible. Is that possible here?

edg2s (changes requested):

test/fixtures/yaml/invalid.yaml
Pchelolo: Done

jdforrester (approved):

jdforrester:
Good from my POV; Ed?

edg2s:
Do we want to turn off `max-len` like we do in JSON? If so you will need to update the hack in `mediawiki.js` too: https://github.com/wikimedia/eslint-config-wikimedia/commit/247ed9d7af35902fd703c501822f0b3d9416be25

Pchelolo:
> Do we want to turn off `max-len` like we do in JSON? If so you will need to update the hack in `mediawiki.js` too: [247ed9d](https://github.com/wikimedia/eslint-config-wikimedia/commit/247ed9d7af35902fd703c501822f0b3d9416be25)

Yaml supports multiline string blocks for any really long strings. Also, the intent of using YAML instead of JSON is human-readability, so I'd hope we will not have configs aimed at being human-readable and with 20 levels of nesting in them. So I'd say let's try to live with max-len and disable it if it creates unavoidable troubles?

edg2s (approved):

PR #432, opened 2022-01-05
vue: Enforce component-name-in-template-casing

### LINKED ISSUE #431: Set vue/component-options-name-casing for MediaWiki
eslint-plugin-vue version 8.2.0 introduces the [`vue/component-options-name-casing`](https://eslint.vuejs.org/rules/component-options-name-casing.html) rule. It defaults to `PascalCase` which is what we want for Vue code in general, but for code in MediaWiki we will need to set it to `kebab-case` instead. This is because the way that we use Vue templates in ResourceLoader through an HTML parser suffers from the same syntax incompatibilities as DOM templates, which is why PascalCase doesn't work there, and the Vue style guide [recommends](https://v3.vuejs.org/style-guide/#component-name-casing-in-templates-strongly-recommended) `kebab-case` for DOM templates,

catrope:
In MediaWiki's environment, enforce kebab-case, since using PascalCase
leads to problems. In non-MW environments, enforce PascalCase, as
recommended in the Vue style guide.

Fixes #431.

### DIFF
--- a/mediawiki.js
+++ b/mediawiki.js
@@ -36,7 +36,8 @@ const config = {
 						"normal": "never",
 						"component": "never"
 					}
-				} ]
+				} ],
+				"vue/component-name-in-template-casing": [ "error", "kebab-case" ]
 			}
 		}
 	]
--- a/test/fixtures/mediawiki/invalid.vue
+++ b/test/fixtures/mediawiki/invalid.vue
@@ -19,6 +19,8 @@
 		<p v-i18n-html:foo />
 		<!-- eslint-disable-next-line vue/html-self-closing, vue/no-unregistered-components -->
 		<blah-component />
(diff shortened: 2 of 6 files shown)


jdforrester:
Normally we'd call this commit 

> vue: Enforce `component-name-in-template-casing` with value `kebab-case`

Is that OK?

jdforrester (dismissed):

catrope:
I'm OK with the format of that proposed commit message, but it's not accurate. We're only setting it to `kebab-case` in the MediaWiki-specific rules. For the non-MW Vue rules, we're setting it to `PascalCase`.

catrope:
How about this? (Force-pushed to change the commit message, same commit otherwise)

jdforrester (approved):

PR #433, opened 2022-01-05
Update eslint-plugin-vue to 8.2.0

catrope:
Add rules:
- vue/no-child-content
    - In MediaWiki, add v-i18n-html as an additional directive
- vue/no-expose-after-await
- vue/prefer-separate-static-class

### DIFF
--- a/test/fixtures/mediawiki/valid.vue
+++ b/test/fixtures/mediawiki/valid.vue
@@ -17,6 +17,10 @@
 		<!-- Valid: vue/component-name-in-template-casing -->
 		<blah-component></blah-component>
 		<!-- Valid: vue/no-child-content -->
+		<!--
+			This line asserts valid code for the no-child-content rule, but there's no way
+			to do that without running afoul of the no-v-html rule
+		-->
 		<!-- eslint-disable-next-line vue/no-v-html -->
 		<p v-html="foo"></p>
 		<p v-i18n-html:foo></p>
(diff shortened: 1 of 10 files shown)


jdforrester (dismissed):

test/fixtures/mediawiki/valid.vue
```diff
@@ -14,8 +14,11 @@
 			</template>
 		</blah-component>
 		<!-- Valid: vue/html-self-closing -->
-		<p v-i18n-html:foo></p>
 		<blah-component></blah-component>
+		<!-- Valid: vue/no-child-content -->
+		<!-- eslint-disable-next-line vue/no-v-html -->
+		<p v-html="foo"></p>
```
edg2s, line 25: This should go in invalid.vue?

edg2s (changes requested):

test/fixtures/mediawiki/valid.vue
catrope: Confusingly, this is asserting that a tag that uses a directive (`v-html`) and is empty is correct under the `vue/no-child-content` rule. However, there's no way to test that without also running afoul of the `vue/no-v-html` rule. The corresponding case in `invalid.vue` is something like `<p v-html="foo">bar</p>`.

edg2s: Comment this then?

edg2s (approved):

PR #445, opened 2022-02-08
build: Add missing yaml.json to package files

edg2s:

PR #446, opened 2022-02-08
tests: Check that all files are included in package.json's files list

edg2s:
This will avoid mistakes like #445, which isn't the first time.


edg2s:
This should fail tests until #445 is merged.

jdforrester (approved):

PR #448, opened 2022-02-09
common: Add stylelint- comments to max-len ignorePattern

catrope:
Lines with // eslint-disable comments are already ignored.
However, in .vue files, ESLint runs the max-len rule against
the <style> block too for some reason, so it complains about
line lengths for stylelint-disable comments.

Add stylelint- to the regex for disabling max-len, so that we don't
complain about stylelint-disable comments being too long.

### DIFF
--- a/common.json
+++ b/common.json
@@ -43,7 +43,7 @@
 		"max-len": [ "warn", {
 			"code": 100,
 			"tabWidth": 4,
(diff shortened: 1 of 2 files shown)


Volker-E (dismissed):

edg2s (dismissed):
Add a test case

jdforrester (approved):

PR #453, opened 2022-04-20
mediawiki: Get browsers list from browserslist-config-wikimedia

edg2s:
This should be a no-op change.


edg2s:
See also https://github.com/wikimedia/stylelint-config-wikimedia/pull/165

jdforrester (approved):

PR #460, opened 2022-04-27
vue: Update eslint-plugin-vue to 8.7.1

edg2s:


edg2s:
@catrope should consider if there are any new rules we should enable: https://github.com/vuejs/eslint-plugin-vue/releases

jdforrester (approved):

catrope:
I've looked through the new rules and I don't think there are any that we should add (except for [prefer-import-from-vue](https://eslint.vuejs.org/rules/prefer-import-from-vue.html), but that has already been added to the default set, so we don't need to add it ourselves).

PR #466, opened 2022-05-17
vue: Use Vue 3 by default

catrope:
This changes the client-es6 and vue-es6 rule sets to use vue3-es6
instead of vue2-es6, making Vue 3 the default for ES6 code.

ES6 code needing to use Vue 2 rules can manually combine the
wikimedia/client-common and wikimedia/vue2-es6 rules.

ES5 code still uses Vue 2 rules. We don't have a vue3-es5 rule set,
because Vue 3 requires ES6. Code targeting Vue 3 should use the ES6
rules.

### DIFF
--- a/client-es6.json
+++ b/client-es6.json
@@ -2,6 +2,6 @@
 	"extends": [
 		"./client-common",
 		"./language/es6",
-		"./vue3-es6"
+		"./vue-es6"
 	]
 }
--- a/client-es6.json
+++ b/client-es6.json
@@ -2,6 +2,6 @@
 	"extends": [
 		"./client-common",
 		"./language/es6",
-		"./vue2-es6"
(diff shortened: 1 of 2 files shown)


client-es6.json
```diff
@@ -2,6 +2,6 @@
 	"extends": [
 		"./client-common",
 		"./language/es6",
-		"./vue2-es6"
+		"./vue3-es6"
```
AnneTee: Question from someone who's not very familiar with this repository: why not just extend `vue-es6` and specify the version in `vue-es6.json` only? I would think that would be a little bit more future proof since you'd only be defining the version in a single place.

catrope: Hmm that's a good idea, I'll do that

jdforrester (approved):

PR #469, opened 2022-05-21
mediawiki: Upgrade eslint-plugin-mediawiki to v0.4.0

jdforrester:
* New rule: `no-nodelist-unsupported-methods` (Ed Sanders)
* New rule: `no-extended-unicode-identifiers` (Roan Kattouw)
* Rule fix: `msg-doc`: Fix short URL to rule help page (Sam Wilson)
* Rule fix: `class-doc`: Fix linting of classList.toggle (Ed Sanders)

### DIFF
--- a/mediawiki.js
+++ b/mediawiki.js
@@ -19,6 +19,8 @@ const config = {
 	"rules": {
 		"mediawiki/class-doc": "error",
 		"mediawiki/msg-doc": "error",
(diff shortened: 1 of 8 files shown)


edg2s (dismissed):
You need to enable the new rules as there is no preset config (maybe there should an "all" config).

catrope (approved):

PR #472, opened 2022-05-31
build: Run tests before linting

edg2s:


Volker-E:
Could you expand? Isn't linting computational cheaper and should therefore come first?

jdforrester:
Ed has made this change in a bunch of repos on the basis that he thinks our linters are style-only and don't point out meaningful failures that test cases won't catch first. I don't really agree, but those who merge his patches clearly do.

edg2s:
> on the basis that he thinks our linters are style-only and don't point out meaningful failures that test cases won't catch first.

That's not the reason (although test failures are typically more critical than lint failures, and so should usually be considered higher priority, but they are not always "style-only" or not "meaningful").

I think Timo made this change first somewhere else. When you are developing you want to unit test code with style errors. In the current configuration you can't run unit tests without having perfectly formatted code first. Perfect formatting is something you typically do last, once things are working.

Quicker linting feedback can be achieved by configuring an IDE plugin, or by running `npm run lint`.

Krinkle (approved):

PR #482, opened 2023-01-11
upgrade "eslint-plugin-jsdoc" to 39.6.4

lectrician1:
The overall package does not support Node 18/19 because of the outdated version of the "eslint-plugin-jsdoc" dependency. This upgrades the version.


edg2s:
Note this is just a warning, the linter still works.

jdforrester:
Done in #538 and others instead.

PR #484, opened 2023-01-11
jsdoc: Update eslint-plugin-jsdoc to 39.2.2

edg2s:


edg2s:
39.2.3 updates some library that causes all our tests to fail - needs more investigation, but this should resolve https://github.com/wikimedia/eslint-config-wikimedia/issues/510

jdforrester (approved):

PR #487, opened 2023-01-16
language: Replace es-x/no-array-prototype-keys/values with no-restricted-syntax

### LINKED ISSUE #488: Fix false positives for Object.keys() and foo.values;
@matmarex wrote:
> I was also negatively surprised by `es-x/no-array-prototype-keys` and `es-x/no-array-prototype-values`. I learned about the changes when I noticed noticed all three being disabled in https://gerrit.wikimedia.org/r/c/mediawiki/extensions/UploadWizard/+/838537.
> 
> In particular, `es-x/no-array-prototype-keys` is triggered by `Object.keys( foo )`, which is obviously not calling the method being disallowed.
> 
> Back in the day, when we first disallowed some methods in [57b2fdc](https://github.com/wikimedia/eslint-config-wikimedia/commit/57b2fdc5ae47af43f4cebc36fb00ea7a9b1cf815), we were actually smart enough to disallow `foo.keys()` without disallowing `Object.keys( foo )`: [57b2fdc#diff-994979481b5e2ff41767c0b7f3485267ce7402a2c0f9a66945ff2785959927a0R88-R91](https://github.com/wikimedia/eslint-config-wikimedia/commit/57b2fdc5ae47af43f4cebc36fb00ea7a9b1cf815#diff-994979481b5e2ff41767c0b7f3485267ce7402a2c0f9a66945ff2785959927a0R88-R91) so I'm hopeful that this kind of solution can be restored.

edg2s:
Fixes #488

### DIFF
--- a/language/es5.json
+++ b/language/es5.json
@@ -9,6 +9,19 @@
 		"ecmaVersion": 5
 	},
 	"rules": {
-		"es-x/no-array-prototype-find": "off"
+		"es-x/no-array-prototype-find": "off",
+		"es-x/no-array-prototype-keys": "off",
+		"es-x/no-array-prototype-values": "off",
+		"no-restricted-syntax": [
+			"error",
+			{
+				"selector": "CallExpression[callee.type='MemberExpression'][callee.property.type='Identifier'][callee.property.name='keys'][callee.object.name!='Object']",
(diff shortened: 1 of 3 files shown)


MatmaRex:
Yes please

jdforrester (approved):

PR #490, opened 2023-02-06
common: Also inherit from plugin:security/recommended

jdforrester:
Fixes https://phabricator.wikimedia.org/T328568.

### DIFF
--- a/common.json
+++ b/common.json
@@ -1,6 +1,7 @@
 {
 	"extends": [
 		"eslint:recommended",
+		"plugin:security/recommended",
 		"./jsdoc"
 	],
 	"env": {},
@@ -20,7 +21,7 @@
 			"extends": "./yaml"
 		}
 	],
-	"plugins": [ "unicorn" ],
+	"plugins": [ "security", "unicorn" ],
 	"rules": {
 		"array-bracket-spacing": [ "error", "always" ],
 		"array-callback-return": "error",
@@ -122,10 +123,11 @@
 		} ],
 		"switch-colon-spacing": [ "error", { "after": true, "before": false } ],
 		"unicode-bom": [ "error" ],
+		"wrap-iife": "error",
+		"yoda": [ "error", "never" ],
+		"security/detect-object-injection": "off",
 		"unicorn/prefer-date-now": "error",
 		"unicorn/prefer-string-slice": "error",
-		"unicorn/throw-new-error": "error",
-		"wrap-iife": "error",
-		"yoda": [ "error", "never" ]
+		"unicorn/throw-new-error": "error"
 	}
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,7 @@
 				"eslint-plugin-no-jquery": "^2.7.0",
 				"eslint-plugin-node": "^11.1.0",
 				"eslint-plugin-qunit": "^7.3.0",
+				"eslint-plugin-security": "^1.7.1",
 				"eslint-plugin-unicorn": "^42.0.0",
 				"eslint-plugin-vue": "^8.7.1",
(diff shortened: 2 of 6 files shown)


sbassett29 (dismissed):
LGTM, thanks.  I'm hoping eslint-plugin-security doesn't become too noisy, but I suppose we can see how it performs on a few more js codebases.

jdforrester:
> LGTM, thanks. I'm hoping eslint-plugin-security doesn't become too noisy, but I suppose we can see how it performs on a few more js codebases.

I think `security/detect-object-injection` is likely to be the noisiest in general. In a test run on OOUI, VE, and MW core, I got 449, 416, and 1053 new warnings, of which 408, 396, and 992 are just this one. 

In practice, LibraryUpgrader will switch all repos over, downgrading errors to warning as appropriate, but it may create a lot of noise across the codebase estate, hiding other warnings.

sbassett29:
@edg2s - Any chance of a review/merge for this?  Should be fairly straight-forward, thanks.

edg2s:
> > LGTM, thanks. I'm hoping eslint-plugin-security doesn't become too noisy, but I suppose we can see how it performs on a few more js codebases.
> 
> I think `security/detect-object-injection` is likely to be the noisiest in general. In a test run on OOUI, VE, and MW core, I got 449, 416, and 1053 new warnings, of which 408, 396, and 992 are just this one.
> 
> In practice, LibraryUpgrader will switch all repos over, downgrading errors to warning as appropriate, but it may create a lot of noise across the codebase estate, hiding other warnings.

This seems like too much warning noise. I'd rather we start with a smaller subset of more reliable rules.

sbassett29:
> This seems like too much warning noise. I'd rather we start with a smaller subset of more reliable rules.

Ok.  If we removed the `security/detect-object-injection` rule, I think that would likely address your concerns.  I do think there is some value here in running these security checks in CI once noisier rules are eliminated.

edg2s:
The repo's description describe this plug as "ESLint rules for Node Security", do they have much value in the client?

sbassett29:
> The repo's description describe this plug as "ESLint rules for Node Security", do they have much value in the client?

So the aforementioned `security/detect-object-injection` rule, while perhaps too noisy and fairly low-risk most of the time, IMO, is absolutely relevant for both NodeJS and client-side JS.  One can see the other rules here and on github:

https://www.npmjs.com/package/eslint-plugin-security

While a couple are fairly Node-specific (expressjs, etc.) the ones dealing with bi-directional characters, eval, regexps, etc. are certainly relevant to both NodeJS and client-side JS.

sbassett29:
I don't have perms to modify PRs for this repo, but disabling `security/detect-object-injection` should look something like the attached patch file.  I think this would eliminate most of the noise from `eslint-plugin-security`, while preserving some valuable security rules.  It would be really nice if we could merge this and at least see how it performs in CI.
[pr490update.patch](https://github.com/wikimedia/eslint-config-wikimedia/files/10972028/pr490update.patch)

jdforrester:
> I don't have perms to modify PRs for this repo, but disabling `security/detect-object-injection` should look something like the attached patch file. I think this would eliminate most of the noise from `eslint-plugin-security`, while preserving some valuable security rules. It would be really nice if we could merge this and at least see how it performs in CI. [pr490update.patch](https://github.com/wikimedia/eslint-config-wikimedia/files/10972028/pr490update.patch)

Oh, sorry about that; I've tweaked and re-based the patch. Does this look good to you to land?

sbassett29:
> Oh, sorry about that; I've tweaked and re-based the patch. Does this look good to you to land?

No problem, lgtm!

catrope (approved):

PR #491, opened 2023-02-24
build: Ensure eslint-plugin-qunit 7.3.0 or later

Krinkle:
The bug fix at https://github.com/platinumazure/eslint-plugin-qunit/pull/231 was released in March 2022 as part of 7.3.0.

Fresh installs of eslint-config-wikimedia already expand `^7.2.0` to 7.3.4, but I'd rather set a minimum so that the next propagation of eslint-config-wikimedia will naturally update this, because otherwise npm is conservative nowadays and only updates when forced to or when doing a full rebuild after deleting one's lock file.


jdforrester:
Not a build change, this is a real dependency. :-)

jdforrester (approved):

PR #492, opened 2023-03-27
client: Change default from ES5 to ES6

jdforrester:
This change matches that for MediaWiki 1.41. Code that explicitly wishes to support ES5 clients still can extend wikimedia/client-es5, as has been recommended for a while.

### DIFF
--- a/client.json
+++ b/client.json
@@ -1,5 +1,5 @@
 {
 	"extends": [
-		"./client-es5"
+		"./client-es6"
 	]
 }


Krinkle:
LGTM. I believe lib-up runs on REL branches, is that right? If so, we'll want to "substitute" client-es5 there to avoid quietely introducing ES6 syntax there.

Krinkle (approved):

jdforrester:
> LGTM. I believe lib-up runs on REL branches, is that right? If so, we'll want to "substitute" client-es5 there to avoid quietely introducing ES6 syntax there.

Done in https://gerrit.wikimedia.org/r/c/labs/libraryupgrader/+/670560 :-)

edg2s:
You need to update the README as well.

PR #493, opened 2023-04-04
mediawiki: Use only ES6 browsers for compat/compat

catrope:
MediaWiki no longer supports non-ES6 browsers, so stop raising compat warnings for browsers like IE11


jdforrester (approved):

edg2s:
Shouldn't this be fixed upstream instead?

PR #496, opened 2023-04-11
Revert "mediawiki: Use only ES6 browsers for compat/compat"

edg2s:
Reverts wikimedia/eslint-config-wikimedia#493

This should be fixed upstream instead.

### DIFF
--- a/mediawiki.js
+++ b/mediawiki.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const commonRules = require( './common' );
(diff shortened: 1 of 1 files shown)


catrope:
So long as upstream distinguishes between `modern` and `modern-es6-only`, we should use the latter here. I see you've changed upstream to make those two the same; once we do a release of the upstream package and pull it through here, I'd be OK with merging this change.

catrope:
Superseded by #497

PR #508, opened 2023-06-09
Require empty line between ES6 class members

MatmaRex:
I think this has been a long-standing convention for our ES5 code, and now we can enforce it for ES6 code.

### DIFF
--- a/language/rules-es6.json
+++ b/language/rules-es6.json
@@ -5,6 +5,7 @@
 		"arrow-parens": "error",
 		"arrow-spacing": "error",
 		"default-param-last": "error",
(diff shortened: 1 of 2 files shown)


catrope (approved):

catrope (changes requested):
The tests fail because an example needs to be added to one of the fixtures files. I suggest adding an example of incorrect code (with an eslint-disable-next-line comment) in tests/fixtures/server/invalid.js

MatmaRex:
Not sure if that works… I don't have this repo set up locally :)

catrope (approved):

PR #513, opened 2023-06-27
Upgrade various eslint-plugins to latest

jdforrester:


edg2s:
Fixes #515

server.json
```diff
@@ -3,5 +3,8 @@
 		"./common",
 		"./node",
 		"./language/es2018"
-	]
+	],
+	"rules": {
+		"es-x/no-hashbang": "off"
```
edg2s, line 8: You already turned this off in node.json...

test/fixtures/client-es5/invalid.js
```diff
@@ -140,9 +140,8 @@
 	// not-es6
 	// eslint-disable-next-line es-x/no-array-prototype-includes
 	[].includes();
-	// eslint-disable-next-line es-x/no-array-prototype-entries, es-x/no-object-entries
```
edg2s, line 143: Shouldn't we need to test this rule still, like we do with no-array-protoype-includes?

edg2s (changes requested):

server.json
jdforrester: Yeah, but for some reason I can't remember it didn't work until I did it here too. Possibly a config load order issue?

test/fixtures/client-es5/invalid.js
jdforrester: I believe the rule upstream has been improved to work out it's not contextually an array?

Krinkle: I don't see any obvious changes upstream, but probably happened indirectly.

There is still one test case above in this file to keep the rule covered:

```js
// eslint-disable-next-line es-x/no-array-prototype-entries
[].entries();
```

Krinkle (dismissed):

server.json
Krinkle: Indeed, this "off" instruction counter acts rules loaded by `/language/es2018`. The node.json file doesn't load this rule, so there is no reason for it to turn it off, either.

Maybe it's useful to keep it set in both, to ease customisations where downstream users use `common + node + es<different>`  instead of `server`. However, that would not work for the same reason it doesn't work here, namely that `es<different>` would override it. To avoid confusion, I'll remove the entry in node.json for now.

test/fixtures/client-es5/invalid.js
jdforrester: Perhaps. Thanks!

jdforrester:
LGTM. Note that we can't get rid of the last remaining `npm audit` issue until eslint-plugin-compat lands a fix for its semver dependency (no Issue, but two PRs in flight so hopefully…).

edg2s (approved):

PR #521, opened 2023-07-28
Assert that valid fixtures contain no disables

edg2s:
Valid fixtures should contain only valid code. Adding an inline
disable should not be allowed.

In rare edge cases we need to trigger an error to assert that
another rule is valid, in which case allow an override.

### DIFF
--- a/test/fixtures/mediawiki/valid.vue
+++ b/test/fixtures/mediawiki/valid.vue
@@ -21,7 +21,7 @@
 			This line asserts valid code for the no-child-content rule, but there's no way
 			to do that without running afoul of the no-v-html rule
 		-->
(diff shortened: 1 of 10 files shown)


test/fixtures/mediawiki/valid.vue
```diff
@@ -21,7 +21,7 @@
 			This line asserts valid code for the no-child-content rule, but there's no way
 			to do that without running afoul of the no-v-html rule
 		-->
-		<!-- eslint-disable-next-line vue/no-v-html -->
+		<!-- eslint-disable-next-line vue/no-v-html !allowdisable -->
```
jdforrester, line 24: Joy.

jdforrester (approved):

PR #523, opened 2023-07-28
Avoid variables named with Hungarian notation

edg2s:


jdforrester (approved):
This feels familiar. :-)

PR #536, opened 2023-11-11
jsdoc: add @ignore, @internal and @stable as definedTags

MusikAnimal:
This will be used to indicate the stability of JS methods, following the introduction of the frontend stable interface policy: https://w.wiki/87wa

See https://www.mediawiki.org/wiki/Topic:Xrs17s867zgd9j7b

### DIFF
--- a/jsdoc.json
+++ b/jsdoc.json
@@ -39,6 +39,16 @@
 	},
 	"rules": {
 		"jsdoc/check-param-names": [ "warn", { "allowExtraTrailingParamDocs": true } ],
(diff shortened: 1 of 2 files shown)


edg2s (changes requested):
Add test cases to fixtures?

MusikAnimal:
> Add test cases to fixtures?

Sure thing! I've added the three tags so that the tests make use of them, and thus prove the tags are valid. I'm not sure if this is satisfactory enough, or if there's more I could do with respect to the tests.

edg2s (approved):

PR #543, opened 2023-11-13
node: Update eslint-plugin-n to 16.3.1

edg2s:


edg2s:
This plugin dropped Node 14 support in v16, but this actually breaks our CI

jdforrester:
> This plugin dropped Node 14 support in v16, but this actually breaks our CI

Difference between dropping claimed support and actually breaking? Should we revert from 16.x to 15.x, or just not do this?

jdforrester (approved):

PR #556, opened 2024-02-06
client/common: Disable es-x/no-resizable-and-growable-arraybuffers, t…

### LINKED ISSUE #555: Disable es-x/no-resizable-and-growable-arraybuffers
User methods called `resize` are common. `ArrayBuffer#resize` from ES2024 is unlikely to be used.

This rule is going to trigger too many false positives.

jdforrester:
…oo many false positives

Fixes #555.

### DIFF
--- a/test/fixtures/client/common/valid.js
+++ b/test/fixtures/client/common/valid.js
@@ -1,5 +1,6 @@
 ( function () {
-	function checkHash( input, cachedValue ) {
+	// Off: es-x/no-resizable-and-growable-arraybuffers (#555)
+	function resize( input, cachedValue ) {
 		var hash = JSON.stringify( input );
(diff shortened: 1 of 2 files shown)


edg2s (approved):

test/fixtures/client/common/valid.js
```diff
@@ -1,11 +1,12 @@
...
+	function resize( input, cachedValue ) {
 		var hash = JSON.stringify( input );
 		// Off: security/detect-possible-timing-attacks (#503)
 		if ( hash === cachedValue ) {
 			return true;
 		}
 	}
 
-	checkHash();
+	resize();
```
edg2s, line 11: This would never trigger the warning, as no language rules are loaded in this fixture.

PR #566, opened 2024-04-19
language: Move es-x/no-resizable-and-growable-arraybuffers to rules-es5

### LINKED ISSUE #555: Disable es-x/no-resizable-and-growable-arraybuffers

edg2s:
Fixes: #555

### DIFF
--- a/language/rules-es5.json
+++ b/language/rules-es5.json
@@ -1,5 +1,6 @@
 {
 	"rules": {
-		"es-x/no-symbol-prototype-description": "off"
+		"es-x/no-symbol-prototype-description": "off",
+		"es-x/no-resizable-and-growable-arraybuffers": "off"
 	}
 }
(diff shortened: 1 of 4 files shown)


language/rules-es5.json
```diff
@@ -1,5 +1,6 @@
 {
 	"rules": {
-		"es-x/no-symbol-prototype-description": "off"
+		"es-x/no-symbol-prototype-description": "off",
+		"es-x/no-resizable-and-growable-arraybuffers": "off"
```
jdforrester, line 4: Should this not be error for es5 and off for es6?

edg2s: .resize is not available in all versions < ES2020 or something, so upstream sets it to "error" in those, however we need to turn it off, so we set it "off" in our lowest-supported language (ES5) so it propagates through to all languages.

jdforrester (approved):

PR #573, opened 2024-05-17
languages: Prefer arrow callbacks in ES6

### LINKED ISSUE #572: Prefer arrow callbacks in ES6 code
We should consider the following rules in ES6 code:
* [prefer-arrow-callback](https://eslint.org/docs/latest/rules/prefer-arrow-callback)
* [arrow-body-style](https://eslint.org/docs/latest/rules/arrow-body-style)
* [implicit-arrow-linebreak](https://eslint.org/docs/latest/rules/implicit-arrow-linebreak)

This results in arrow functions being used for callbacks (unless they reference `this` or `arguments`, in which case they would need be converted manually). When arrow functions contain a single return statement, the body-less style is prefered (e.g. `( x ) => x + 1`. Also a single line is prefered for such body-less functions.

edg2s:
Also prefer single-line body-less style when possible.

Fixes #572

### DIFF
--- a/language/rules-es6.json
+++ b/language/rules-es6.json
@@ -15,6 +15,9 @@
 		"no-var": "error",
 		"one-var": "off",
 		"vars-on-top": "off",
-		"prefer-const": "error"
+		"prefer-const": "error",
+		"prefer-arrow-callback": "error",
+		"implicit-arrow-linebreak": "error",
+		"arrow-body-style": "error"
 	}
 }
--- a/test/fixtures/mediawiki/qunit/invalid.js
+++ b/test/fixtures/mediawiki/qunit/invalid.js
@@ -1,6 +1,6 @@
 QUnit.module( 'example' );
 
-QUnit.test( 'test', function ( assert ) {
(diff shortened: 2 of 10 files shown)


edg2s:
cc @matmarex

Krinkle:
If there isn't one already, I suggest adding a `QUnit.test()` case that involves `this.something` (e.g. from a beforeEach hook), to make sure it neither warns about nor tries to "fix" those. Or, I suppose, really in any input file, it doesn't have to be a QUnit test per-se.

edg2s:
prefer-arrow-callback is conservative and the autofix is designed to never introduce an error, e.g. if `this` or `arguments` is present in the function body:

https://eslint.org/docs/latest/rules/prefer-arrow-callback#:~:text=unbound%20function%20expression

edg2s:
Added https://github.com/wikimedia/eslint-config-wikimedia/pull/573/files#diff-86bcbe79dd5ebf49dae82bbc6b83b36dfacc16943ed661dd6d319337a93007a0

Krinkle (approved):

PR #594, opened 2024-11-05
server: Change target from ES2018 (Node 10) to ES2022 (Node 18)

jdforrester:


test/fixtures/server/invalid.js
```diff
@@ -59,43 +59,10 @@
...
-
-	// not-es2020
-	// eslint-disable-next-line es-x/no-string-prototype-replaceall
-	'abc'.replaceAll( 'a', 'x' );
-	// eslint-disable-next-line es-x/no-promise-any
-	Promise.any( [] );
-
-	// not-es2021
-	// Nothing yet
+	// TODO: <!-- eslint-disable-next-line n/no-unsupported-features/node-builtins -->
```
edg2s: are there no new builtins in 20/22?

edg2s: Maybe WebSocket? https://github.com/eslint-community/eslint-plugin-n/commit/73e2bed2e76dc9382069268954ae894665f18538#diff-d4704b37b332ca78aeeebf924fbf92f33ffa2c9c088fc9d112fa02f81c64aa32R23

jdforrester: `error  'WebSocket' is not defined  no-undef` …

jdforrester: Remember, this is ES2022, not Node 22 (~= ES2025)

edg2s: That should still work, it just needs to be Node>18. Either way, something from that list should allow us to test the rule...

jdforrester: The list you've linked is of things new-in-Node-22. This code raises us **to** Node 18, so Node <= 18, please. :-) I suppose we could test for `DOMException`?

edg2s: I know, but the file the change is in has other version numbers.

> DOMException?

Whatever works.

test/fixtures/server/valid.js
```diff
@@ -56,6 +56,45 @@
...
+	// Off: es-x/no-string-prototype-replaceall
+	'abc'.replaceAll( 'a', 'x' );
+	// Off: es-x/no-promise-any
+	Promise.any( [] );
+
+	// ES2021
+	// Nothing yet
+
+	// ES2022
+	const e = new DOMException();
```
edg2s: ..this isn't triggering an error then.

We want to find something that still triggers the /node=builtins error...

edg2s (dismissed):

test/fixtures/server/valid.js
jdforrester: There's nothing in https://github.com/mysticatea/eslint-plugin-node/blob/master/lib/rules/no-unsupported-features/node-builtins.js that starts from Node 23 yet, though.

edg2s: Wrong repo: https://github.com/eslint-community/eslint-plugin-n/blob/master/lib/rules/no-unsupported-features/node-builtins.js

Which loads data from various files. Searching for `"23`: https://github.com/search?q=repo%3Aeslint-community%2Feslint-plugin-n%20%2223&type=code

Krinkle:
Ping from https://gerrit.wikimedia.org/r/c/mediawiki/core/+/1008074.

I'd like to allow use of optional chaining. This landed back in Node 14, but trips a linter in CI even on Node, because it was spec'ed in ES2020, whereas we lint against ES2018.

Krinkle (commented):
I've updated the patch to use `URL.canParse()` in server/invalid.js which is listed in Ed's link as supported since Node 19, thus a good example of something the lint rule should catch as invalid.

`crypto.verify()` is a fine example to move over from invalid.js to valid.js as new in Node 15 which was previously invalid but is now valid. 

I've also added `new Blob()` as a second valid thing, new in Node 18.

Krinkle (dismissed):

server.json
```diff
@@ -2,9 +2,10 @@
 	"extends": [
 		"./common",
 		"./node",
-		"./language/es2018"
+		"./language/es2022"
 	],
 	"rules": {
+		"n/no-unsupported-features/node-builtins": [ "error", { "version": ">18.0.0" } ],
```
edg2s: This should read from package.json's "engines" field (although I note we have a type and call in "engine"), instead of being defined in the config.

jdforrester: We don't use the `engines` field in the real world. I vaguely understand it to be deprecated?

edg2s: It isn't deprecated. We may not use it but this plugin does. We also need to change the override in the test's eslintrc.json settings key, which is other way to set it.

edg2s: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines

jdforrester: "npm devs: Why specify structured data as such, instead of a randomly parsed string?"

Krinkle: It's not deprecated, but its use is generally limited to the uncommon use case where 1) a repo is publishing a package to npm, and 2) it has local and project-specific reasons that require a newer Node version than otherwise (e.g. not merely the max of dependencies engines minimum + wikimedia minimum).

We inherit browser compat requirements and ES version from the ESLint present, instead of hardcoding in every single repo separately. Even for server code, we have effectively set the Node.js minimum already in the wikimedia preset, by proxy of defining an ES version for the "server" preset. But we currently have a gap in linting, where it only looks for ES APIs implemented by the Node.js version, and not the Node.js runtime APIs. Providing this by default in same way sounds useful to me. In particular because (per the above) we don't have a practice of setting `engines` everywhere. Starting such practice now would create ambiguity (was it a substituted default? Or an intentional local override?).

edg2s (approved):

PR #596, opened 2024-12-02
jquery: Enable no-done-fail

jdforrester:
Raise eslint-plugin-no-jquery to 3.1.0+ to provide new rules.

### DIFF
--- a/jquery.json
+++ b/jquery.json
@@ -12,6 +12,7 @@
 		"no-jquery/no-animate-toggle": "error",
 		"no-jquery/no-class-state": "error",
 		"no-jquery/no-constructor-attributes": "error",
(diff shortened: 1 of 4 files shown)


Volker-E (dismissed):

edg2s:
Why no-fx? We don't use jQuery slim anywhere or enforce rules relating to it. We already disallow the deprecated parts of $.fx.

jdforrester:
> Why no-fx? We don't use jQuery slim anywhere or enforce rules relating to it. We already disallow the deprecated parts of $.fx.

Only because it was new. Happy to only enable `no-done-fail`.

edg2s (approved):

PR #599, opened 2024-12-19
client: Change target from ES2015 (ES6) to ES2016

siddharthvp:
As all mainstream browsers passing the current MediaWiki feature checks support ES2016, raise the default wikimedia/client configuration to allow ES2016 syntax and methods.

Bug: T381537


Volker-E:
Seems reasonable to me. In context, we just moved Codex to ES2016, see https://phabricator.wikimedia.org/T382391

siddharthvp:
Fixed test failure.

jdforrester (dismissed):

jdforrester:
To confirm, this is waiting on sign-off on https://phabricator.wikimedia.org/T381537 before I'll merge it.

Krinkle:
@jdforrester That task is about the next Grade A milestone, which is currently proposing ES2017. That is something different from this PR, which is correcting our **current** milestone of ES2016.

Our current Grade A requirement, as moved to by me in https://phabricator.wikimedia.org/T178356, went from ES5 to ES7 (ES2016), and our current startup.js requirements already reflect that. Likewise, the syntax checker for Gadgets and user scripts is also set to ES7/ES2016 already (<https://gerrit.wikimedia.org/r/c/mediawiki/core/+/962739>). Now, 99% of the features people talk about came in ES6, so communication was largely centred on that. In fact the difference was so small that I actually could not find a feature test that reliably checked for ES6 without letting ES7 in. And indeed, if you review `startup.js` with the two features in mind that ES7/ES2016 adds (Array-includes and `**` operator), these all have caniuse.com data with browser versions lower than our current minimum.

Krinkle (approved):

PR #604, opened 2025-02-18
Remove ES5 configs

### LINKED ISSUE #603: Delete ES5 configs
These shoud be unused by Wikimedia projects, and are a maintenance burden.

edg2s:
Fixes #603


NovemLinguae:
Is this a PR to delete the ruleset wikimedia/client-es5? I still see a bunch of uses of this in codesearch.

https://codesearch.wmcloud.org/deployed/?q=es5&files=eslintrc%5C.json%24#mediawiki/extensions/GlobalBlocking

Perhaps we should replace all those with wikimedia/client (which I think auto picks the latest supported mediawiki es version) before proceeding with this PR.

Or do a silent redirect of wikimedia/client-es5 to wikimedia/client. But that could be confusing to users expecting ES5.

jdforrester (approved):

jdforrester:
> Is this a PR to delete the ruleset wikimedia/client-es5? I still see a bunch of uses of this in codesearch.
> 
> https://codesearch.wmcloud.org/deployed/?q=es5&files=eslintrc%5C.json%24#mediawiki/extensions/GlobalBlocking

Yeah, that's fine, releases of CI tools are almost always by their nature breaking changes, which this will be; migrations of users will have to take this into account. LibUp already tries to automatically do this, so repos still on the old config have something broken about them already, or were missed in a manual migration.

edg2s:
Most of those repos using client-es5 are still using ES5 style code (e.g. `var` over `const`/`let`). We should probably tell libup to skip them until upgraded.

edg2s:
...or we can have libup switch them over to `wikimedia/client` and then do it's usual thing of converting any errors to warnings, so they can be fixed at a later date.

edg2s:
See https://gitlab.wikimedia.org/repos/ci-tools/libup/-/merge_requests/70

PR #607, opened 2025-02-19
mediawiki: set "sourceType" to "commonjs" and enable "no-implicit-globals"

anny21:
Bug: [T367441](https://phabricator.wikimedia.org/T367441)

### DIFF
--- a/mediawiki/common.js
+++ b/mediawiki/common.js
@@ -23,12 +23,14 @@ const config = {
 			browsers.join( ',' )
 		]
 	},
+	"parserOptions": {
+		"sourceType": "commonjs"
+	},
 	"overrides": [
 		{
 			"files": [ "**/*.vue" ],
 			"extends": [ "plugin:mediawiki/vue" ],
 			"rules": {
-				"no-implicit-globals": "off",
 				"vue/html-self-closing": [ "error", {
 					"html": {
 						"void": "never",


jdforrester:
You're changing it here for the code in this repo, but I imagine you instead want to change it in the 'client' config that users of this library?

Krinkle:
I suggest `mediawiki/common` instead, as we also have Wikimedia projects with standalone frontend code outside MediaWiki. This guruantee is something that MediaWiki/ResourceLoader provides, similar to how Node.js guruantees it for server code.

When we set `sourceType: commonjs`, I suggest we remove the `no-implicit-globals` rule at the same time. That rule was added in https://github.com/wikimedia/eslint-config-wikimedia/pull/225 as a compromise for Vue modules and Package modules, but as @edg2s mentioned there, that rule is useful even in ESM/Vue/CJS module scope. It currently disables too much, but it was all we could do back then.

Now that debug mode v1 is gone, all MediaWiki code is guruanteed to have a commonjs module scope. This was already a guruantee for all code in production mode, and in debug=2 mode, and for Package module code it was guruanteed even in debug mode v1. But, there remained one narrow case: a traditional script module loaded in debug mode v1. That scenario no longer exists, which means <https://phabricator.wikimedia.org/T50886> is finally upon us, and the ESLint preset can now reflect this.

When `sourceType: commonjs`, the `no-implicit-globals` rule is okay to leave enabled (we can remove the "off" override), because it will know that `var foo;` in a file does not create a global variable.

* [ESLint Playground: sourceType script and no-implicit-globals enabled  (warning)](https://eslint.org/play/#eyJ0ZXh0IjoidmFyIGZvbyA9IDE7Iiwib3B0aW9ucyI6eyJydWxlcyI6eyJuby1pbXBsaWNpdC1nbG9iYWxzIjpbImVycm9yIl19LCJsYW5ndWFnZU9wdGlvbnMiOnsicGFyc2VyT3B0aW9ucyI6eyJlY21hRmVhdHVyZXMiOnt9fSwiZWNtYVZlcnNpb24iOjIwMTYsInNvdXJjZVR5cGUiOiJzY3JpcHQifX19)
* [ESLint Playground: sourceType commonjs and no-implicit-globals enabled  (no warning)](https://eslint.org/play/#eyJ0ZXh0IjoidmFyIGZvbyA9IDE7Iiwib3B0aW9ucyI6eyJydWxlcyI6eyJuby1pbXBsaWNpdC1nbG9iYWxzIjpbImVycm9yIl19LCJsYW5ndWFnZU9wdGlvbnMiOnsicGFyc2VyT3B0aW9ucyI6eyJlY21hRmVhdHVyZXMiOnt9fSwiZWNtYVZlcnNpb24iOjIwMTYsInNvdXJjZVR5cGUiOiJjb21tb25qcyJ9fX0=)

anny21:
Thank you, @jdforrester and @Krinkle! Updated the patch! Please have a look!

edg2s (approved):

jdforrester (approved):

PR #611, opened 2025-02-25
vue: Delete vue2 configs

### LINKED ISSUE #606: Drop support for Vue2
Our Vue config is quite complicated, but presumably we only want to support one (major) version of Vue on MW, at least going forward. It would be nice if we could simplify our config by dropping Vue2 support.

edg2s:
Fixes #606


package.json
```diff
@@ -41,17 +41,14 @@
 		"selenium.json",
 		"server.json",
 		"vue-common.json",
```
jdforrester, line 43: Should we drop everything down to just vue3.json?

jdforrester: Given we just added vue/es2017, no.

jdforrester (approved):

PR #617, opened 2025-05-13
client: Create client/es2017 config

edg2s:


edg2s:
This will soon be the default, but that can be done in another commit.

jdforrester (approved):

PR #623, opened 2025-06-27
Add config for TypeScript files

micgro42:
This takes over the config from https://github.com/wmde/eslint-config-wikimedia-typescript/ which was explicitly only every intended to be a provisional set of configuration.

This temporary configuration led to problems when its version of this configuration here had to be kept up-to-date or automatic upgrades with LibraryUpgrader would fail, which happened regularly.

This adds the configuration here as a `.js` file, because that allows us to retain the comments in the original file, which are still meaningful. Precedence for `.js` config files already exists in the vue3/ directory.

@typescript-eslint/eslint-plugin and @typescript-eslint/parser are added as direct dependencies in accordance with all the other dependencies that are also direct dependencies like `eslint-plugin-qunit` and so one.

Bug: T301431

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+Next release
+===================
+* Add config for TypeScript, copied from WMDE's typescript eslint config (Michael Große)
+
 0.30.0 / 2025-05-22
 ===================
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,3 @@
-Next release
-===================
-* Add config for TypeScript, copied from WMDE's typescript eslint config (Michael Große)
-
 0.30.0 / 2025-05-22
 ===================
 
--- a/common.json
+++ b/common.json
@@ -19,18 +19,8 @@
 			"extends": "./yaml"
 		},
 		{
-			"files": [
-				"**/*.ts"
-			],
-			"extends": [
-				"./typescript.js"
-			],
-			"rules": {
-				"comma-dangle": [
-					"error",
-					"always-multiline"
-				]
-			}
+			"files": [ "**/*.ts" ],
+			"extends": "./typescript"
 		}
 	],
 	"plugins": [ "security", "unicorn" ],
--- a/common.json
+++ b/common.json
@@ -17,6 +17,20 @@
 		{
 			"files": [ "**/*.yaml", "**/*.yml" ],
 			"extends": "./yaml"
+		},
+		{
+			"files": [
+				"**/*.ts"
+			],
+			"extends": [
+				"./typescript.js"
+			],
+			"rules": {
+				"comma-dangle": [
+					"error",
+					"always-multiline"
+				]
+			}
 		}
 	],
 	"plugins": [ "security", "unicorn" ],
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
 		"eslint-plugin-yml": "^1.14.0"
 	},
 	"devDependencies": {
+		"@stylistic/eslint-plugin": "^3.1.0",
 		"escape-string-regexp": "^4.0.0",
 		"fs-readdir-recursive": "^1.1.0",
 		"qunit": "^2.24.1"
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
 		"qunit.json",
 		"selenium.json",
 		"server.json",
+		"typescript.js",
 		"yaml.json",
 		"client/",
 		"language/",
@@ -52,6 +53,8 @@
 	],
 	"license": "MIT",
 	"dependencies": {
+		"@typescript-eslint/eslint-plugin": "^8.0.0",
+		"@typescript-eslint/parser": "^8.0.0",
 		"browserslist-config-wikimedia": "^0.7.0",
 		"eslint": "^8.57.0",
 		"eslint-plugin-compat": "^4.2.0",
--- a/package.json
+++ b/package.json
@@ -53,8 +53,8 @@
 	],
 	"license": "MIT",
 	"dependencies": {
-		"@typescript-eslint/eslint-plugin": "^8.0.0",
-		"@typescript-eslint/parser": "^8.0.0",
+		"@typescript-eslint/eslint-plugin": "8.35.1",
+		"@typescript-eslint/parser": "8.35.1",
 		"browserslist-config-wikimedia": "^0.7.0",
 		"eslint": "^8.57.0",
 		"eslint-plugin-compat": "^4.2.0",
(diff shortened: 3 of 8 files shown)


readme.md
```diff
@@ -119,6 +119,38 @@ The `wikimedia/server` config consists of `wikimedia`, `wikimedia/node` and `wik
...
+config only for TypeScript files like so:
+
+`.eslintrc.json`:
+```json
+{
+	"root": true,
+	"extends": [
+		"wikimedia/server"
+	],
+	"overrides": [
```
edg2s: These overrides should be in common.json so that all .ts files are linted with this config automatically.

edg2s (changes requested):

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+Next release
```
jdforrester: We normally write this as part of the release process, to make sure we don't miss things. Let's drop it from this PR>

package.json
```diff
@@ -52,6 +53,8 @@
 	],
 	"license": "MIT",
 	"dependencies": {
+		"@typescript-eslint/eslint-plugin": "^8.0.0",
```
jdforrester: Is upstream likely to introduce new rules/changed behaviours in 8.1.x etc.? If so, that'd be a breaking change for our CI, so these should be exactly pinned (like we do with eslint-plugin-jsdoc).

micgro42: It has the following in their section titled "A change to the plugins *shall not* be considered breaking if it:"

> Adds additional checks to an existing rule that causes new reports in a small-to-medium set of cases in an average codebase.

([source](https://typescript-eslint.io/users/versioning#eslint-plugin))

Let's pin this (and `@typescript-eslint/parser`, as I understand it, they need to be in sync) exactly.

readme.md
micgro42: This will automatically opt in all typescript files (unless the eslint command is restricted to only some file-types).

But looking at the repos that use TypeScript (https://codesearch.wmcloud.org/things/?q=import&files=.*%5C.ts%24&excludeFiles=&repos=), this does not seem too bad. Wikibase is already using this config, and so are Growth repositories (GrowthExperiments, CommunityConfiguration, GuidedTours).

I'll move the overrides as suggested.

micgro42: Done.

CHANGELOG.md
micgro42: Done.

package.json
micgro42: Done.

micgro42:
For now, I have implemented your feedback in extra commits (as usual for GitHub PRs). If you prefer, I can also squash them so that this PR is only dealing with a single commit, like Gerrit.

common.json
```diff
@@ -19,6 +19,20 @@
 			"files": [ "**/*.yaml", "**/*.yml" ],
 			"parser": "yaml-eslint-parser",
 			"extends": "./yaml"
+		},
+		{
+			"files": [
+				"**/*.ts"
+			],
+			"extends": [
+				"./typescript.js"
```
edg2s: The `.js` is not necessary here.

Maybe use the same formatting as the other overrides above (use one line for files and extends)

common.json
```diff
@@ -19,6 +19,20 @@
...
+			],
+			"rules": {
+				"comma-dangle": [
```
edg2s: We can just put this rule in `typescript.js` - like we do with the other overrides (e.g. `json.json`)

typescript.js
```diff
@@ -0,0 +1,57 @@
...
+
+module.exports = {
+	extends: [
+		'plugin:@typescript-eslint/recommended'
+	],
+	plugins: [
+		'@typescript-eslint'
+	],
+	parserOptions: {
+		parser: '@typescript-eslint/parser',
```
edg2s: "parser" should be at the root level, not inside "parserOptions"

edg2s (changes requested):
Needs some test fixtures to assert that this is at least wired up correctly.

typescript.js
micgro42: Mh, I'm actually not sure about this. IIRC, then this is used in conjecture with vue files that are compiled before being committed and that have a TypeScript `<script>` tag. And then the root-level parser uses `vue-eslint-parser` and the one in parserOptions uses `@typescript-eslint/parser`.

But it has been ages that I worked with such a setup. Let's put it into the root and figure this case out when it arises.

Also, https://typescript-eslint.io/packages/parser/ does not specify any of the existing parser options, so I dropped them all as well.

common.json
micgro42: Done.

common.json
micgro42: Done

typescript.js
micgro42: ... or at least that is what I thought. It seems that at least 
```
		ecmaFeatures: {
			impliedStrict: true
		}
```
has an effect? I've added those options back. Clearly, I don't fully understand how they work and where they are documented yet.

micgro42:
> Needs some test fixtures to assert that this is at least wired up correctly.

Done. Though the `valid.ts` file feels a bit pointless.

typescript.js
```diff
@@ -0,0 +1,59 @@
...
+		'@typescript-eslint/array-type': [ 'error', { default: 'array' } ],
+		'@typescript-eslint/explicit-function-return-type': [ 'error', {
+			allowExpressions: true,
+			allowTypedFunctionExpressions: true,
+			allowHigherOrderFunctions: true
+		} ],
+		'@typescript-eslint/explicit-module-boundary-types': [ 'error' ],
+		'@typescript-eslint/explicit-member-accessibility': [ 'error', { accessibility: 'explicit' } ],
+		// aligned to https://github.com/wikimedia/eslint-config-wikimedia/blob/master/common.json#L21
+		'@stylistic/indent': [ 'error', 'tab', { SwitchCase: 1 } ],
```
edg2s, line 31: We aren't using @stylistic in common yet so we should probably switch over at the same time for consistency.

You could also copy these over dynamically by calling `require( 'common' )` and selectively copying keys.

micgro42: Mh, not sure I want to add that to this PR. Rather, this PR should remain focused on adding configuration for TypeScript. We can either "prepend" another PR that makes use of `@stylistic` in other eslint config, or do that as a follow-up PR. But doing that here too feels like a lot of scope creep for me.

What do you think?

edg2s: I suppose if this is the status quo in eslint-config-wikimedia-typescript we can minimise the migration cost.

edg2s (approved):

jdforrester (approved):

PR #628, opened 2025-10-08
[BREAKING CHANGE] Update dependencies and Node version to >=20

edg2s:


jdforrester:
Can you drop the CI config for Node 18 in this too, so it doesn't have the ugly errors?

jdforrester (dismissed):

jdforrester:
In the CHANGELOG we'll need to (tediously) list all the renamed rules. Fun.

jdforrester (approved):

PR #643, opened 2025-11-19
jsdoc: Update to 61.3.0

edg2s:


jdforrester:
Looking at https://github.com/gajus/eslint-plugin-jsdoc/compare/v61.0.0...v61.3.0 the four new rules are covered by your patch, so no need to flag (or make a major release).

jdforrester (approved):

PR #647, opened 2025-11-27
build: Create a release script

edg2s:


jdforrester:
Eh. The "hard" bit is deciding what bits go in what section, and reading each commit title to add in the new/changed rules when they don't mention it. I'm not sure this script will increase easy/speed of anything meaningfully, but we can merge if you insist.

jdforrester (approved):

edg2s:
This means all you have to do is edit the changelog, rather than build it from scratch with the right version (and update package.json etc).

PR #655, opened 2026-04-29
build: Fix release.sh to honor preferred names via mailmap

Krinkle:
Use `%aN` instead of `%an` which honors the mailmap. This matches what we already do in npm and composer packages hosted in Wikimedia Gerrit via commands like `composer changelog` [1]

This removes the need to manually expand Ed's name, since this is in the mailmap already.

Document the rest of the release process while at it.

[1] https://codesearch.wmcloud.org/search/?q=%22changelog%22&files=%5E%28composer.json%7Cpackage.json%29&excludeFiles=&repos=

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
 	"scripts": {
 		"test": "qunit test/test.js && npm run lint",
 		"lint": "eslint . --max-warnings 0",
+		"changelog": "git log HEAD --format=\"* %s (%aN)\" --no-merges --not --tags | sort",
 		"build-invalid": "node test/test.js dump"
 	},
 	"repository": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -6,6 +6,7 @@
 	"scripts": {
 		"test": "qunit test/test.js && npm run lint",
 		"lint": "eslint . --max-warnings 0",
+		"changelog": "git log HEAD --format=\"* %s (%aN)\" --no-merges --not --tags | sort | grep -vE \"^\\* (build|docs?|tests?):\"",
```
jdforrester: This removal is not documented, and not wanted?

jdforrester (changes requested):

jdforrester (approved):

PR #661, opened 2026-07-08
jquery: Upgrade eslint-plugin-no-jquery to 6.0.0; raise target deprecations to jQuery 4.0

jdforrester:
Done as two commits in case you think the 4.0 jump is premature (even if it's a no-op).

### DIFF
--- a/jquery.json
+++ b/jquery.json
@@ -22,6 +22,7 @@
 		"no-jquery/no-global-selector": "error",
 		"no-jquery/no-grep": "error",
 		"no-jquery/no-in-array": "error",
+		"no-jquery/no-jquery-ui": "error",
 		"no-jquery/no-map-util": "error",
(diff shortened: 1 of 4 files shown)


edg2s:
The jump to jQuery 4 is not strictly a no-op - it upgrades deprecation warnings to error.

We can also consider enabling `no-jquery-ui`. Any project that still uses jqui can disable it.

jdforrester:
> The jump to jQuery 4 is not strictly a no-op - it upgrades deprecation warnings to error.

Fair. Do you want me to split this into two MRs, then?

> We can also consider enabling `no-jquery-ui`. Any project that still uses jqui can disable it.

We could! That should probably be its own MR, however. Edit: Pushed as a third commit into this stack; we'll need to remember to rebase/merge not squash.

Krinkle (approved):

PR #665, opened 2026-07-27
Release v0.32.5

Krinkle:


jdforrester (commented):
I think this should be 0.33.0 as there are a few new CI rules which makes this a breaking change.

jdforrester (approved):

